PR #4, opened 2019-06-28
Card generator

fentontaylor:
Add CardGenerator class that reads text file and generates cards from it.
The make_cards method can handle text files that use different separators, which are passed as an argument to make_cards.
flashcard_runner now uses CardGenerator to create the cards for the deck/round.


fentontaylor:
oops sorry, I thought I was on my repo

PR #240, opened 2021-07-11
Move game operation to new file

TannerDale:
Move game operation from flashcard runner to game for ease of testing


TannerDale:
Meant to be local, please ignore

PR #246, opened 2021-07-11
Iteration 3 Complete

chloell5:
All four lib/ files complete as per specifications with each method tested in spec/
flashcard_runner.rb runs as outlined.

Still to do:
Format for invalid inputs
Iteration 4


chloell5:
Closed for no name in PR title

PR #247, opened 2021-07-11
Chloe Price

chloell5:
Iteration 3 complete
All four lib/ files complete as per specifications with each method tested in spec/
flashcard_runner.rb runs as outlined.

Still to do:
Format for invalid inputs
Iteration 4


chloell5:
Accidental duplicate

PR #275, opened 2021-10-13
Number correct

Eldridge-Turambi:
Good start to #take_turn method but cannot figure out how to go to the next card of @deck after it is called. Please work on that.


Eldridge-Turambi:
Wrong branch sorry about that

PR #283, opened 2021-10-15
For mentor code review

joel-grant:


lib/flashcard_runner.rb
```diff
@@ -0,0 +1,33 @@
...
+cards = [card_1, card_2, card_3, card_4]
+deck = Deck.new(cards)
+round = Round.new(deck)
+
+
+puts
+puts "Welcome! You're playing with #{cards.size} cards."
+puts "-------------------------------------------------"
+puts "This is card number #{round.turns.size + 1} out of #{cards.size}."
+
```
dmeskis: Ha, this works fine for now but I was referring to github comments @joeldenverdev

lib/round.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  def current_card
+    @deck.cards[0]
+  end
+
+  def take_turn(guess)
+    turn = Turn.new(guess, @deck.cards[0])
+    @turns << turn
+    @deck.cards.shift()
```
dmeskis, line 16: Look into what `.shift` returns, maybe you can simplify this method

lib/round.rb
```diff
@@ -0,0 +1,62 @@
...
+      end
+    end
+    return count
+
+  end
+
+  def number_correct_by_category(category)
+    count = 0
+    @turns.each do |turn|
+      if turn.category == category  # @dmeskis rspec gives an error that there is an
```
dmeskis: I can't pull this down to test myself right now, but i definitely don't see `turn#category` defined. To debug, maybe throw in a pry and check what `turn.card` returns. Inspect the class, and you can also see what methods are defined on the objects with `turn.card.methods`. If it is indeed an instance of `Card`, `attr_accesor` should be defining that method for you. I can look more in depth later but those are a couple debugging tips you could use for now

lib/round.rb
dmeskis: Actually, I would be wary of using `shift` here. My hunch is you are seeing some side effects from modifying the cards array object that is causing the problems you are seeing in the runner.

dmeskis: `@deck.cards` here points to the same object that the `cards` variable in the runner file points to. Think about how the destructive method `shift` might affect the `cards.each` loop in the runner on line 22

PR #329, opened 2021-12-04
Iteration 3

jennerj:
This PR incorporates all of iteration 3, wrote 3 extra cards and got all text to print out to user properly.


jennerj:
creation!

PR #366, opened 2022-05-11
Jonathan Pope

jonathanmpope:


jonathanmpope:
Jonathan Pope - final pull request

lib/card.rb
```diff
@@ -0,0 +1,10 @@
+class Card
+    attr_reader :question, :answer, :category
+
+  def initialize (question, answer, category)
+    @question = question
+    @answer = answer
+    @category = category
+  end
+
+end
```

lib/card_generator.rb
```diff
@@ -0,0 +1,27 @@
...
+    array = @data.map do |card|
+        card.split(",")
+    end
+    array.map do |array|
+       question = array.shift
+       answer = array.shift.downcase 
+       category = array.shift
+       Card.new(question, answer, category)
+    end
+  end
```
megstang, line 25: really nice work here!

lib/deck.rb
```diff
@@ -0,0 +1,22 @@
...
+  category_cards = []
+  @cards.each do |card|
+    if card.category == category
+      category_cards << card
+    end
+  end
+  category_cards
+  end
+
+end
```
megstang, line 22: 🎉  nice!

lib/round.rb
```diff
@@ -0,0 +1,49 @@
+class Round
+  attr_reader :deck, :turns, :index, :number_correct, :correct_array, :category_array
```
megstang, line 2: You only need attr_readers for attributes that you want to read outside of your class. For that reason, the only attr_readers that you need to have are deck, turns, and number_correct.

lib/round.rb
```diff
@@ -0,0 +1,49 @@
...
+    @card_deck = deck.cards
+  end
+
+  def current_card
+    @card_deck[0]
+  end
+
+  def take_turn(guess)
+    another_turn = Turn.new(guess, current_card)
+    @category_array << another_turn.card.category
```
megstang, line 21: you do already have a method for cards_in_category that give you back the cards in a particular category. For that reason, I don't think it's necessary that you create this variable. Similarly, on line 43, when you call to this variable, you could just call to the method that returns the cards in a particular category and count them up.

lib/turn.rb
```diff
@@ -0,0 +1,27 @@
+class Turn
+    attr_reader :card, :guess, :feedback
```
megstang, line 2: you have a method called `feedback`, so no need for this here. You do not need an attr_reader for all attributes -- you only need attr_readers for attributes you want to read when outside of the class.

megstang: In fact, your handwritten method is scoped such that ruby will run your handwritten method first before looking at the attr_Reader method for feedback, so this is never used.

lib/turn.rb
```diff
@@ -0,0 +1,27 @@
...
+end
+
+def feedback
+  if correct? == true
+    @feedback = 'Correct!'
+  else
+    @feedback = 'Incorrect.'
+  end
+  @feedback
+end
```
megstang, line 25: Going with the suggestion above, I'd suggest removing the attrreader AND the attribute `@feedback`, and just let this method return a string: 

```
def feedback
  if correct? == true
    return 'Correct!'
  else
    return 'Incorrect.'
  end
end
```

You're storing unnecessary state by having an attribute for feedback.

spec/round_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+
+  it 'will be able to read the deck variable' do
+
+    expect(@round.deck).to eq @deck
+  end
+
+  it 'will be able to take turns' do
+    @round.turns
+
+    expect(@round.turns).to eq []
```
megstang, line 29: this happens a lot in your tests -- you do not need to run a method in order to assert it on your expectation line. AKA, line 27 is unnecessary.

megstang (commented):
Nice work! I left a few comments for you to look over. Let me know if you have any questions. I'll follow up in slack with your scores before the end of the day.

PR #368, opened 2022-05-11
Eli Sachs

easachs:
Iteration 1-4 complete

### DIFF
--- a/lib/round.rb
+++ b/lib/round.rb
@@ -1,0 +1,3 @@
+class Round
+
+end
--- a/lib/round.rb
+++ b/lib/round.rb
@@ -30,12 +30,20 @@ class Round
   end
 
   def percent_correct
-    @number_correct.to_f * 100 / @turns.count
+    if @turns.count == 0
+      return "Cannot divide by 0"
+    else
+      @number_correct.to_f * 100 / @turns.count
+    end
   end
 
   def percent_correct_by_category(cat)
-    number_correct_by_category(cat) * 100 / @turns.count do |turn|
-      turn.card.category == cat
+    if @turns.count == 0
+      return "Cannot divide by 0"
+    else
+      number_correct_by_category(cat) * 100 / @turns.count do |turn|
+        turn.card.category == cat
+      end
     end
   end
 
--- a/lib/round.rb
+++ b/lib/round.rb
@@ -1,26 +1,25 @@
 class Round
 
   attr_reader :deck,
+              :current_card,
               :used_cards,
               :turns,
               :number_correct
 
   def initialize(deck)
     @deck = deck
+    @current_card = @deck.cards.first
     @used_cards = []
     @turns = []
     @number_correct = 0
   end
 
-  def current_card
-    @deck.cards.first
-  end
-
   def take_turn(guess)
-    @turns << Turn.new(guess, @deck.cards.first)
-    @number_correct += 1 if @turns.last.guess == @deck.cards.first.answer
-    @used_cards << @deck.cards.first
+    @turns << Turn.new(guess, current_card)
+    @number_correct += 1 if @turns.last.guess == current_card.answer
+    @used_cards << current_card
     @deck.cards.shift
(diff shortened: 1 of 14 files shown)


easachs:
I am curious how this could be adapted to work with commas in the questions of the cards.txt files (ex: "What county is Centennial, CO in?")

lib/round.rb
```diff
@@ -0,0 +1,79 @@
...
+    puts "-" * 30
+
+    total_cards = @deck.cards.length + @used_cards.length
+    card_number = 0
+
+    while card_number < total_cards
+      card_number = @used_cards.length + 1
+      puts "This is card number #{card_number} out of #{total_cards}."
+      puts "Question: #{@current_card.question}"
+      answer = gets.chomp.to_s.downcase
```
abdulredd, line 61: Just want to point out that there is no reason to convert the response you get using `gets` into a string because it comes back as a string by default, even if the user enters in numbers.

```ruby
answer = gets.chomp.downcase
# answer.is_a?(String) # => true
```

lib/round.rb
```diff
@@ -0,0 +1,79 @@
...
+      take_turn(answer)
+      puts @turns.last.feedback
+      if !@turns.last.correct?
```
abdulredd, line 64: Keep in mind that you wrote your `take_turn` method to `return` the last `turn` so on line **62** you could have stored your call to `take_turn` in a variable and use that variable for lines **63** & **64**

```ruby
turn = take_turn(answer)
puts turn.feedback
if !turn.correct?
#[REST OF CODE...]
```

lib/round.rb
```diff
@@ -0,0 +1,79 @@
...
+        puts "The correct answer was #{@turns.last.card.answer}."
+      end
+    end
+
+    puts "***** Game over! *****"
+    puts "You had #{@number_correct} correct guesses out of #{total_cards} for a total score of #{percent_correct.to_i}%."
```
abdulredd, line 70: I would recommend converting a float type to an integer type because you run the risk of losing the decimals. 

```ruby
33.333333.to_i # => 33
```

I would much rather you keep the decimals and specify how many decimal places you want to retain

```ruby
33.333333.round(1) # => 33.3
```

This would obviously have to be clarified against the specifics of the project as to whether decimals should be displayed or dropped like you have it.

lib/round.rb
```diff
@@ -0,0 +1,79 @@
...
+    used_cards_by_category = @used_cards.group_by do |card|
```
abdulredd, line 71: Awesome use of creating a hash here to store your cards by category keys. Our hash class came after this project was submitted so great job thinking out of the box!

lib/round.rb
```diff
@@ -0,0 +1,79 @@
...
+      card.category
+    end
+    stats = used_cards_by_category.each do |k,v|
```
abdulredd, line 74: Iterating over a hash FTW!

lib/round.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    while card_number < total_cards
+      card_number = @used_cards.length + 1
+      puts "This is card number #{card_number} out of #{total_cards}."
+      puts "Question: #{@current_card.question}"
+      answer = gets.chomp.to_s.downcase
+      take_turn(answer)
+      puts @turns.last.feedback
+      if !@turns.last.correct?
+        puts "The correct answer was #{@turns.last.card.answer}."
```
abdulredd, line 65: Love how you display the correct answer when the entered guess is incorrect!

lib/turn.rb
```diff
@@ -0,0 +1,18 @@
...
+    @guess = guess
+    @card = card
+  end
+
+  def correct?
+    @guess.downcase == @card.answer.downcase
+  end
+
+  def feedback
+    self.correct? ? "Correct!" : "Incorrect."
```
abdulredd, line 15: `self` isn't required here but it's good to see that you know how to use it. Also great use case for using a ternary statement.

PR #369, opened 2022-05-12
First Pull Request

nicecodingtoo:
Created four files with defined classes, as of yet, only the turn_spec passes.


nicecodingtoo:
Completed and passed tests for files: card, turn, and deck.

nicecodingtoo:
Final pull request.

flashcard_runner.rb
```diff
@@ -0,0 +1,84 @@
...
+  # #
+  # # category - (@round.percent_correct_by_category(category)).to_i
+  # # percentage.to_1
+  # #
+  # # puts "#{percentage.to_1}"
+  # # end
+  #
+
+
+  guess = gets.chomp
```
cjsim89, line 52: I think you were working through the runner file when you turned this in - currently this does not work as the spec intends, though it does output all of this to the terminal. You'll want to set up variables and use your class instances to run the game.

lib/deck.rb
```diff
@@ -0,0 +1,23 @@
...
+    @cards = cards
+  end
+
+  def count
+    @cards.count
+  end
+
+  def cards_in_categories(category)
+    cards_in_categories = []
+  @cards.each  do |card|
```
cjsim89, line 14: Careful of your indents in this method; it's a little hard to read as is.

lib/round.rb
```diff
@@ -0,0 +1,45 @@
+class Round
+  attr_reader :deck, :turns
+
+  def initialize(deck)
+    @deck = deck
+    @turns = []
+  end
+
+  def current_card
+    deck.cards[turns.count]
```
cjsim89, line 10: Be careful when referencing instance variables that you use the `@` sign consistently.

lib/round.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def take_turn(guess)
+    new_turn = Turn.new(guess, current_card)
+    @turns << new_turn
+    new_turn
+  end
+
+   def number_correct
+     @turns.count do |turn|
```
cjsim89, line 20: Did you intend to use `.each` here instead of `.count`...?

lib/turn.rb
```diff
@@ -0,0 +1,27 @@
...
+    guess == card.answer
+  end
+
+#This method either returns "Correct!" or "Incorrect."
+#based on whether the guess was correct or not.
+  def feedback
+    return "Correct!"
+  end
+
+  def feedback2
```
cjsim89, line 23: This method `feedback2` is not in the spec, and I think you were trying to have two different methods return separate feedback. However, both sets of feedback (both Correct and Incorrect) should be able to be returned from the same `feedback` method. 

What kind of syntax could you use to determine a course of action in code, based on a condition?

spec/card_spec.rb
```diff
@@ -1,4 +1,4 @@
-require './lib/card'
+require './lib/card.rb'
```
cjsim89, line 1: You can just use `./lib/card`, you should not need an `.rb`. in these require statements.

PR #375, opened 2022-05-13
Michael Bonini

mkbonini:
Finished Iteration 4.

### DIFF
--- a/lib/card.rb
+++ b/lib/card.rb
@@ -1,0 +1,8 @@
+class Card
+  attr_reader :question, :answer, :category
+  def initialize(question, answer, category)
+    @question = question
+    @answer = answer
+    @category = category
+  end
+end
--- a/lib/card_generator.rb
+++ b/lib/card_generator.rb
@@ -10,8 +10,4 @@ class CardGenerator
        @cards << Card.new(card_input[0],card_input[1],card_input[2].chomp)
      end
   end
-
-  def make_cards
-
-  end
 end
--- a/lib/card_generator.rb
+++ b/lib/card_generator.rb
@@ -1,0 +1,17 @@
+class CardGenerator
+  #require './cards.txt'
+  attr_reader :cards
+  def initialize(filename)
+    @filename = filename
+    @cards = []
+    @card_data = File.readlines(File.join(__dir__, filename))
+    @card_data.each do |line|
+       @card_input = line.split(",")
+       @cards << Card.new(card_input[0].chomp,card_input[1].chomp,card_input[2].chomp)
+     end
+  end
+
+  def make_cards
+
+  end
+end
--- a/lib/card_generator.rb
+++ b/lib/card_generator.rb
@@ -1,13 +1,13 @@
 class CardGenerator
-  #require './cards.txt'
   attr_reader :cards
   def initialize(filename)
     @filename = filename
     @cards = []
-    @card_data = File.readlines(File.join(__dir__, filename))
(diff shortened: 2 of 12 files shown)


lib/card.rb
```diff
@@ -0,0 +1,8 @@
+class Card
+  attr_reader :question, :answer, :category
+  def initialize(question, answer, category)
+    @question = question
+    @answer = answer
+    @category = category
+  end
+end
```
megstang, line 8: 👍

lib/card_generator.rb
```diff
@@ -0,0 +1,13 @@
...
+    @filename = filename
+    @cards = []
+    card_input = []
+    card_data = File.readlines(File.join(__dir__, filename))
+    card_data.each do |line|
+       card_input = line.split(",")
+       @cards << Card.new(card_input[0],card_input[1],card_input[2].chomp)
+     end
+  end
+end
```
megstang, line 13: looks great! NIce work

lib/deck.rb
```diff
@@ -0,0 +1,18 @@
...
+  def cards_in_category(category)
+    return_cards = []
+    cards.each do |card|
+      if card.category == category
+        return_cards << card
+      end
+    end
+    return return_cards
+  end
+end
```
megstang, line 18: for syntax convention, I'd just add a line between each of your methods.

lib/round.rb
```diff
@@ -0,0 +1,76 @@
...
+  end
+
+  def current_card
+    return deck.cards[@guess]
+  end
+
+  def take_turn(guess)
+      turn = Turn.new(guess, current_card)
+      @turns << turn
+      @guess += 1
```
megstang, line 18: Since you're incementing guess each time you shovel in a turn object, you may consider removing the need for the guess instance variable since you could just call `@turns.count` whenever you need to know how many guesses there are.

lib/round.rb
```diff
@@ -0,0 +1,76 @@
...
+      if turn.correct? && turn.card.category == category
+        correct += 1
+      end
+    end
+    return correct
+  end
+
+  def percent_correct
+    return number_correct * 100.0 / @guess
+  end
```
megstang, line 44: So, just to reiterate, instead of `@guess` here, you could just call `@turns.count`

lib/round.rb
```diff
@@ -0,0 +1,76 @@
...
+  def percent_correct_by_category(category)
+    return number_correct_by_category(category) * 100.0 / @deck.cards_in_category(category).length
```
megstang, line 46: nice use of cards_in_category method!

lib/round.rb
```diff
@@ -0,0 +1,76 @@
...
+  end
+  def categories_list
+
+    categories_holder = []
+    @deck.cards.each do |card|
+      categories_holder << card.category
+    end
+    return categories_holder.uniq
+
+  end
```
megstang, line 56: 👍

lib/turn.rb
```diff
@@ -0,0 +1,19 @@
+class Turn
+  attr_reader :guess, :card
+  def initialize(guess, card)
+    @guess = guess
+    @card = card
+  end
+
+  def correct?
+    return guess.downcase == card.answer.downcase
```
megstang, line 9: nice edge case consideration here 👍

megstang (commented):
nice work! I left a few comments for you to look over. Let me know if you have any questions! I'll follow up in slack later today with scores.

PR #380, opened 2022-05-14
Brenna's Flashcard Project

brennacodes:
Working as expected.
Complete through Iteration 4.

Added link in Readme to my spinoff of this project, the "[Trivia Machine](https://github.com/brennacodes/trivia_machine)" 😁 It's more fully-featured and interactive. Yay!

### DIFF
--- a/flashcard.rb
+++ b/flashcard.rb
@@ -33,7 +33,6 @@ class FlashcardRunner
       @round.card_complete
       ask_question
     end
-    # game_over
   end
 
   def game_over
--- a/flashcard.rb
+++ b/flashcard.rb
@@ -1,0 +1,46 @@
+require './lib/helper.rb'
+
+class FlashcardRunner
+  def initialize(round, deck, cards)
+    @round = round
+    @deck = deck
+    @cards = cards
+  end
+
+  def start
+    puts "Welcome! You're playing with 4 cards.
+         -------------------------------------------------
+         This is card number #{@round.turn_count + 1} out of 4."
+    ask_question
+  end
+
+  def ask_question
+    puts "Question: #{@round.current_card.question}"
+    check_input
+  end
+
+  def check_input
+    guess = gets.chomp.downcase
+    @round.take_turn(guess)
+    get_feedback
+  end
+
+  def get_feedback
+    puts @round.turn_feedback
+    if @round.turn_count == @deck.cards.size
+      game_over
+    else
+      @round.card_complete
+      ask_question
+    end
+    # game_over
+  end
+
+  def game_over
+    puts "****** Game over! ******
+     You had #{@round.number_correct} correct guesses out of 4 for a total score of #{@round.percent_correct}%.
+     STEM - #{@round.percent_correct_by_category('STEM')}% correct
+     Turing Staff - #{@round.percent_correct_by_category('Turing Staff')}% correct
+     Pop Culture - #{@round.percent_correct_by_category('Pop Culture')}% correct"
+  end
+end
--- a/lib/card_generator.rb
+++ b/lib/card_generator.rb
@@ -1,0 +1,23 @@
+require_relative './helper'
+
+class CardGenerator
+  attr_reader :file, :cards
+
+  def initialize(file)
+    @file = './lib/' + file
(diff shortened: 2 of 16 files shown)


spec/turn_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    expect(turn.guess).to eq("Juneau")
+  end
+
+  it "can check if a guess is correct" do
+    expect(turn.correct?).to eq(true)
+  end
+
+  it "can give the user feedback" do
+    expect(turn.feedback).to eq("Correct!")
+  end
```
epintozzi, line 21: We should also test that it returns the expected result when `correct?` is false and `feedback` is `Incorrect.`

lib/round.rb
```diff
@@ -0,0 +1,65 @@
+require 'set'
+require_relative 'helper'
+
+class Round
+  attr_reader :deck, :turns, :turn_count
+
+  @@rounds = Set.new
```
epintozzi, line 7: This totally works, however it would be a little more conventional to object oriented programing to instead create a new class, perhaps some like `Game` that could contain and manage info about multiple `rounds`.

spec/round_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+    expect(round.deck.cards.count).to eq(3)
+  end
+
+  it 'can return cards from a category' do
+    expect(round.deck.cards_in_category(:STEM)).to eq([card2, card3])
+  end
+
+  it 'can check the current card' do
+    expect(round.current_card).to_not eq(deck.cards.last)
+  end
```
epintozzi, line 25: We would also want to test in the affirmative as well. There are lots of things `round.current_card` does not equal, but we should also verify that it returns the expected value as well:
`expect(round.current_card).to_eq eq(card1)`

spec/round_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  it 'can take a turn' do
```
epintozzi, line 27: We should also test that the `take_turn` method returns a `Turn` object as per the spec. Right now it looks like this method returns an array so a test like this would have helped us catch that. 

```ruby  
it 'returns a turn object when take_turn' do
    expect(round.take_turn('Juneau')).to be_a(Turn)
  end
```

lib/card_generator.rb
```diff
@@ -0,0 +1,21 @@
+require_relative 'helper'
+
+class CardGenerator
+  attr_reader :filename
```
epintozzi, line 4: Since we aren't accessing `filename` outside of this class, we don't need an `attr_reader`

flashcard.rb
```diff
@@ -0,0 +1,45 @@
...
+    ask_question
+  end
+
+  def ask_question
+    puts "Question: #{@round.current_card.question}"
+    check_input
+  end
+
+  def check_input
+    guess = gets.chomp.downcase
```
epintozzi, line 23: Because `answer` on `Card` is not always downcased, this can created some incorrect results. For example, in the `cards.txt` the answer to the cardboard cutout question is "Justin bieber". If the user's answer is downcased, then it will always be stored as "justin bieber" which will always return as incorrect because `guess == answer` will never return as true if the answer isn't also stored as downcase upon creation.

flashcard.rb
```diff
@@ -0,0 +1,45 @@
+require './lib/helper.rb'
+
+class FlashcardRunner
+  def initialize(round, deck, cards)
+    @round = round
+    @deck = deck
+    @cards = cards
+  end
```
epintozzi, line 8: This could and should be instantiated only with `round` because round already knows what `deck` and `cards` are. `deck` is only ever being called directly on line 30 which could easily be revised to `@round.deck.cards.size`

flashcard.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def start
+    puts "Welcome! You're playing with 4 cards.
```
epintozzi, line 11: It would be more dynamic to interpolate with something like `#{@round.deck.cards.count}` instead of hardcoding the number of cards here and on line 13.

PR #381, opened 2022-05-14
Michael Koulouvaris

mikekoul:


lib/deck.rb
```diff
@@ -0,0 +1,16 @@
...
+  def initialize(cards)
+    @cards = cards
+  end
+
+  def count
+    cards.count
+  end
+
+  def cards_in_category(category)
+    cards.select { |card| card.category == category }
```
cjsim89, line 13: Nice job using `.select`

lib/deck.rb
```diff
@@ -0,0 +1,16 @@
+class Deck
+  attr_accessor :cards
```
cjsim89, line 2: Wondering about your thought process to use an `attr_accessor` instead of `attr_reader`...? I'm guessing since you needed to change the value of the cards in the Round class maybe?

lib/deck.rb
```diff
@@ -0,0 +1,16 @@
...
+  attr_reader :cards_in_category
+  def initialize(cards)
+    @cards = cards
+  end
+
+  def count
+    cards.count
```
cjsim89, line 9: Be careful about referencing instance variables without using the `@` sign - it's totally ok and Ruby lets us do it _if_ you're also using an `attr_reader` or `accessor`. It's a good idea to just match them up every time with `@` to keep track of them, especially in bigger more complex programs.

lib/round.rb
```diff
@@ -0,0 +1,65 @@
...
+  attr_reader :deck, :turns, :number_correct, :percent_correct
+  def initialize(deck)
+    @deck = deck
+    @turns = []
+    @number_correct = 0
+    @percent_correct = 0.0
+  end
+
+  def current_card
+    deck.cards[0]
```
cjsim89, line 11: Try to remember to use `@` when referencing instance variables.

lib/round.rb
```diff
@@ -0,0 +1,65 @@
...
+    puts "Question: #{current_card.question}"
+    take_turn
+    puts "This is card number #{turns.count + 1} out of #{deck.cards.count}"
+    puts "Question: #{current_card.question}"
+    take_turn
+    puts "****** Game over! ******"
+    puts "You had #{number_correct} guesses out of #{turns.count} for a total score of #{percent_correct}%"
+    puts "Geography - #{percent_correct_by_category(:Geography)}% correct"
+    puts "STEM - #{percent_correct_by_category(:STEM)}% correct"
+    puts "Pop_Culture - #{percent_correct_by_category(:Pop_Culture)}% correct"
```
cjsim89, line 62: The category names are hard-coded; could you have used the categories from the cards in the beginnings of these strings?

lib/turn.rb
```diff
@@ -0,0 +1,16 @@
+class Turn
+  attr_accessor :guess, :card
```
cjsim89, line 2: Why use an `attr_accessor` here vs. an `attr_reader`?

lib/card.rb
```diff
@@ -0,0 +1,9 @@
+class Card
+  attr_accessor :answer, :category
```
cjsim89, line 2: Curious why you might use an attr_accessor for these two properties specifically, and not the `:question` as well?

lib/deck.rb
mikekoul: My understanding of attr_accessor is still foggy at best so I thought I needed an accessor when it came to class on class interactions. Ill spend more time understanding what it does

lib/deck.rb
mikekoul: Heard. I do it cause I feel it looks cleaner

lib/round.rb
mikekoul: I feel like I tried that and it was giving me problems. So I just went with the hard code method.

mikekoul:
Thank you for the feedback

PR #383, opened 2022-05-14
Mayu Takeda

okayama-mayu:


flashcard_runner.rb
```diff
@@ -0,0 +1,22 @@
...
+# deck = Deck.new([card_1, card_2, card_3])
+# round = Round.new(deck)
+
+# use cards.txt for the flashcards.
+filename = "cards.txt"
+generator = CardGenerator.new(filename)
+deck = Deck.new(generator.cards)
+round = Round.new(deck)
+
+round.start
```
cjsim89, line 22: This is a great runner file - it uses all the code you wrote in classes by requiring them here, and it is short & to the point. Nice job!

lib/round.rb
```diff
@@ -0,0 +1,145 @@
...
+  def correct?
+    round.turns.first.guess == round.turns.fist.card.answer
+  end
+
+  def number_correct
+    num_correct = 0
+    turn_index = 0
+    until turn_index == @turns.count
+      turns.each do |turn|
+        if turns[turn_index].guess == turns[turn_index].card.answer
```
cjsim89, line 30: I see you're directly comparing the guess to the answer - this will determine if the guess _exactly matches_ the answer; an edge case here that should be accounted for is if the case is off. For instance if I type "Justin Bieber", it is no less correct than "justin bieber" which is what you have in your `cards.txt` file.

cjsim89: This is something to keep in mind whenever you are gathering user input - it's a much better experience for the user to allow for capitalizations etc.

lib/round.rb
```diff
@@ -0,0 +1,145 @@
...
+    end
+    num_correct
+  end
+
+  def number_correct_by_category(category)
+    num_correct = 0
+    turn_index = 0
+    until turn_index == @turns.count
+      @turns.each do |turn|
+        if @turns[turn_index].card.category == category && @turns[turn_index].guess == @turns[turn_index].card.answer
```
cjsim89, line 46: Similar to the above comment; comparing direct matches with guesses & answers provides a worse user experience than if we compare against `guess.downcase`, for example.

lib/round.rb
```diff
@@ -0,0 +1,145 @@
...
+      end
+    end
+    num_correct
+  end
+
+  def number_correct_by_category(category)
+    num_correct = 0
+    turn_index = 0
+    until turn_index == @turns.count
+      @turns.each do |turn|
```
cjsim89, line 45: I think this is slightly more code than we need, though it does work! 
You needed to go through the `turn_index` to access the correct turn, so that's why you have `until turn_index == @turns.count` and then you're incrementing the `turn_index` below this. However, you're also doing `@turns.each` - there is a method called `.each_with_index` that provides you the same functionality as `.each`, but a built-in index as well. 
So your method could look like this: 
```
@turns.each_with_index do |turn, index| 
   if turn.card.category == category && #turn guess is correct
   ...
```

cjsim89: In fact, I'm not quite sure that we need an index here...? Maybe we do and I just don't see it. In any case, the code does work, just might be a little too engineered.

lib/round.rb
```diff
@@ -0,0 +1,145 @@
...
+      end
+    end
+    num_correct
+  end
+
+  def percent_correct
+    num_correct = 0
+    turn_index = 0
+    until turn_index == @turns.count
+      @turns.each do |turn|
```
cjsim89, line 61: Similar to above, you may have a few too many loops/iterators here.

lib/round.rb
```diff
@@ -0,0 +1,145 @@
...
+
+  def start
+    card_number = 0
+
+    puts "Welcome! You're playing with #{deck.count} cards.
+    -------------------------------------------------"
+    until card_number == deck.count
+      puts "This is card number #{card_number + 1} out of #{deck.count}.\nQuestion: #{deck.cards[0].question}"
+      guess = gets
+      if guess.strip == @deck.cards[0].answer
```
cjsim89, line 103: You can also do `gets.strip` in order to clean up this code a bit.

cjsim89: The `@deck.cards[0]` is fine, just misleading because of how the array works... Since you're using `.rotate!` on the deck between turns, the `current_card` is always going to be at `[0]` place. Reading the code through the first time, it can be confusing but it does depend on that `rotate` method working.

lib/round.rb
```diff
@@ -0,0 +1,145 @@
...
+      end
+    end
+    cats.uniq
+  end
+
+  # for Iteration 4: iterate through the categories to output the % correct correct by category
+  def category_results
+    cat_index = 0
+    until cat_index == self.category_collector.count
+      self.category_collector.each do |category|
```
cjsim89, line 138: You do not need `self` to access the same class's methods. `self` is a reserved keyword for doing something similar here, but it's called a Class Method and we'll learn about it later.

lib/round.rb
```diff
@@ -0,0 +1,145 @@
...
+        index += 1
+      end
+    end
+    cats.uniq
+  end
+
+  # for Iteration 4: iterate through the categories to output the % correct correct by category
+  def category_results
+    cat_index = 0
+    until cat_index == self.category_collector.count
```
cjsim89, line 137: Again, I think this is over-engineered to use too many loops/indexes. Maybe I am wrong 🤷

lib/turn.rb
```diff
@@ -0,0 +1,20 @@
+class Turn
+  attr_reader :card, :guess 
+
+  def initialize(guess, card)
+    @guess = guess
+    @card = card
+  end
+
+  def correct?
+    guess == @card.answer
```
cjsim89, line 10: I see you're using the `@` to reference instance variables interchangeably - my advice is to **not** do that; if you're going to reference, access, or re-assign any instance variables, do so with the `@`. Otherwise it can get confusing and inconsistent with what you're trying to reference.

cjsim89: In other words, this should look like 

`@guess == @card.answer`

That way, we know instantly that `@guess` refers to line 5, and not a method somewhere else.

spec/card_spec.rb
```diff
@@ -1,27 +1,17 @@
...
-    card = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
-
-    expect(card).to be_instance_of(Card)
-  end
-
-  it 'has a question' do
-    card = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
-
-    expect(card.question).to eq("What is the capital of Alaska?")
+  before(:each) do
```
cjsim89, line 4: Nice job using `before :each` - do you find it saves you time?

spec/card_spec.rb
```diff
@@ -1,27 +1,17 @@
...
-    expect(card.answer).to eq("Juneau")
+  it 'exists' do
+    expect(@card).to be_instance_of(Card)
   end
 
-  it 'has a category' do
-    card = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
-
-    expect(card.category).to eq(:Geography)
+  it 'has a question' do
```
cjsim89, line 12: Your test is described as checking for `has a question`, but your `expect` statements actually do 3 things: 
1. Checks for question 
2. Checks for its answer
3. Checks for its category. 

It's fine to put multiple `expect` statements in one test, but it should always be if the test is describing all 3 things, or makes that test more robust in some way. For instance, you could rename this test to 
`it "has a question, answer, and category" do`
or
`it "has attributes" do`

spec/turn_spec.rb
```diff
@@ -0,0 +1,102 @@
+require './lib/turn'
+require './lib/card'
+
+RSpec.describe Turn do
+  before(:each) do
+    @card = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    @turn = Turn.new("Juneau", @card)
+  end
+
+  describe Card do
```
cjsim89, line 10: Since the `Card` class was tested in `card_spec.rb`, this describe block might not be necessary here.

spec/turn_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+
+  describe "#feedback" do
+    it 'gives you feedback' do
+      expect(@turn.feedback).to eq("Correct!")
+    end
+  end
+
+end
+
+RSpec.describe Turn do
```
cjsim89, line 57: You shouldn't need to describe `Turn` twice in the same file - is there a reason you have two describe blocks for Turn..?

PR #384, opened 2022-05-15
Tyler Caudill

TeeCaud:


lib/deck.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def cards_in_category(category)
+    @cards.find_all do |card|
+      card.category == category
+  end
+end
+
+
+
```
epintozzi, line 18: It's generally convention to remove extra blank lines at the end of files classes.

lib/deck.rb
```diff
@@ -0,0 +1,19 @@
...
+  def initialize(cards = [])
+    @cards = cards
+  end
+
+  def count
+    @cards.count
+  end
+
+  def cards_in_category(category)
+    @cards.find_all do |card|
```
epintozzi, line 12: Nice use of `find_all`!

spec/turn_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+    turn = Turn.new("Juneau", card)
+
+    expect(turn.guess).to eq "Juneau"
+    expect(turn.card).to eq card
+
+  end
+
+  it "it checks if the answer's correct" do
+    card = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    turn = Turn.new("Juneau", card)
```
epintozzi, line 26: This test is failing because the answer you are passing in has a capital J but the `correct?` method in your `Turn` class does a `downcase` on `card.answer` so they do not match. You can either update your test to pass in a lower case answer ("juneau") or update your `correct?` method to also call `downcase` on `guess`.

spec/turn_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  it "it checks if the answer's incorrect" do
+    card = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    turn = Turn.new("Andrew", card)
+
+    expect(turn.correct?).to eq false
+
+  end
+
+  it "gives you correct feedback" do
```
epintozzi, line 40: This test is failing for the same reason as above.

spec/round_spec.rb
```diff
@@ -0,0 +1,170 @@
...
+    round.take_turn("Mardfdes")
+    expect(round.percent_correct_by_category(:STEM)).to be 50.0
+
+
+
+  end
+
+
+
+
```
epintozzi, line 169: There's a lot of extra white space in this spec file. Ideally we only have one blank line in between each assertion.

spec/turn_spec.rb
epintozzi: This is also causing a bunch of tests for `Round` to fail. Addressing the case-sensitive aspect of the guess will resolve all of them though 👍🏻

lib/round.rb
```diff
@@ -0,0 +1,97 @@
...
+      (number_correct_by_category(category) / cards_by_category(category).to_f) * 100
+
+
+  end
+
+  def start
+    # require 'pry'; binding.pry
+    puts "Welcome! You're playing with #{@deck.cards.count} cards."
+    puts "-------------------------------------------------"
+    @deck.cards.count.times do
```
epintozzi, line 67: I like that you've made this dynamic rather than hardcoding that there are 4 cards. This way, we can add as many cards as we want without having to update this method.

epintozzi (commented):
Nice work! I will send scores in Slack by end of day Friday. Let me know if you have any questions on this feedback.

PR #386, opened 2022-05-15
Bryce Simonds

brycesimonds:


lib/deck.rb
```diff
@@ -0,0 +1,37 @@
+class Deck
+  attr_reader :cards, :stem_cards, :geography_cards, :turing_cards, :pop_cards
+  def initialize(cards)
+    @cards = cards
+    @stem_cards = []
+    @geography_cards = []
+    @turing_cards = []
+    @pop_cards = []
+    cards.each do |card|
```
abdulredd, line 9: I completely understand the idea behind what you're doing here but this only accounts for cards that are in these 4 categories. If someone wanted to supply cards that were in any other category, they wouldn't be added to an array.

Having just `@cards = cards` would have been fine, and then to do your `cards_in_category` method, you could have done something like this:

```ruby
  @cards.select do |card|
    card.category == category
  end
end
``` 

This way we loop through the array of cards and only select those cards that match the category we pass in as an argument. Doing it this way, you can have any number of cards of any category and it would still work.

lib/round.rb
```diff
@@ -0,0 +1,121 @@
...
+    @turns = []
+    @number_correct = 0
+    @stem_correct = []
+    @geo_correct = []
+    @turing_correct = []
+    @pop_correct = []
+  end
+
+  def current_card
+    deck.cards[@turn_counter]
```
abdulredd, line 18: Instead of keeping a counter for the number of turn cards, you can also just use the length of the number of items in the array of `turns`

`@turn_counter` will always equal `@turns.length`

```ruby
def current_card
  deck.cards[@turns.length]
end
```

In your `take_turn` method you shovel the `turn` into `@turns` and then increase `@turn_counter` by 1, this second step is unnecessary.

lib/round.rb
```diff
@@ -0,0 +1,121 @@
...
+
+  def percent_correct_by_category(category)
+    if category == :Geography
+      (@geo_correct.count / deck.geography_cards.count).to_f * 100.0
+      elsif category == :STEM
+      (@stem_correct.count / deck.stem_cards.count).to_f * 100.0
+    end
+  end
+
+  def start
```
abdulredd, line 63: I see one of your comments is to refactor this method later so I will help you out with that:

My suggestion would be to use some kind of loop here that keeps track of some sort of count, and stops when a certain condition is met, like when the number of turns at the number of cards in the deck are equal or something similar. That way you aren't doing a lot of hard coding.

```ruby
def start
  counter = 1

  puts "Welcome! You're playing with #{deck.count} cards."
  puts "---------------------------------------------------"

  while counter <= deck.count
    puts "This is card number #{counter} out of #{deck.count}."
    puts  "Question: #{current_card.question}" #<=== Here I'm using our already defined current_card method

    guess = gets.chomp

    turn = take_turn(guess) #<=== Here I'm using our already defined take_turn method
    puts turn.feedback #<=== Here I'm using the feedback method defined on the turn instance object
  end

  puts "You had #{number_correct} correct guesses out of #{@turns.length} for a total score of #{percent_correct}%."

  #[REST OF METHOD...]
end
```

Hopefully this helps shed some light on an approach you could have taken.

lib/round.rb
```diff
@@ -0,0 +1,121 @@
...
+        @pop_correct << guess_4
+      else
+        p "Incorrect."
+      end
+
+      puts "****** Game over! ******"
+      puts "You had #{stem_correct.count + pop_correct.count + geo_correct.count + turing_correct.count} correct guesses out of #{deck.count} for a total score of 75%"
+      puts "STEM - 100% correct"
+      puts "Turing Staff - 50% correct"
+      puts "Pop Culture - 100% correct"
```
abdulredd, line 118: We were not really looking to have these values hardcoded and actually be dynamic based on the actual number of cards answered correctly and the actual category of the cards. Right now as it is coded, no matter what category of cards of used or the number of cards answered correctly or incorrectly, the output will always be the same.

lib/turn.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  end
+
+
+
+  def correct?
+    if @guess == card.answer
+      true
+    elsif @guess != card.answer
+      false
```
abdulredd, line 15: While this code works, and could be used without issue I do want to point out that there is a bit of redundancy here.

All three of the following methods do the exact same thing with less and less code for each one.

```ruby
def correct?
  if @guess == card.answer
    true
  elsif @guess !- card.answer
    false
  end
end
```

```ruby
def correct?
  if @guess == card.answer
    return true
  end

  false
end
```

```ruby
def correct?
  @guess == card.answer #<=== conditional statements return booleans by default, no need to say "if something return true, else return false"
end
```

Again, the method works just fine as written, but just wanted to point it out.

lib/deck.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    if card.category == :STEM
+      @stem_cards << card
+    elsif card.category == :Geography
+      @geography_cards << card
+    elsif card.category == :Turing
+      @turing_cards << card
+    elsif card.category == :Pop
```
abdulredd, line 17: The last statement of any `if/elsif` statement can more often than not just be an `else`

```ruby
if card.category == :STEM
  @stem_cards << card
elsif card.category == :Geography
  @geography_cards << card
elsif card.category == :Turing
  @turing_cards << card
else #<=== If it isn't one of the other options then naturally it has to be this option by default otherwise you risk not having any of the options returning true and possibly getting some unexpected behavior down the line somewhere
  @pop_cards << card
end
```

PR #387, opened 2022-05-15
Blake Saylor

blakesaylor:


lib/deck.rb
```diff
@@ -0,0 +1,23 @@
+class Deck
+
+  attr_reader :cards
+
+  def initialize(cards)
+    @cards = cards
+  end
+
+  def count
+    return @cards.count
```
epintozzi, line 10: Ruby will automatically return the result of the last line of code in a method for us, so we can actually write this as `@cards.count` without the `return`

lib/deck.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  def cards_in_category(category)
+    category_cards = []
+    @cards.each do |card|
+      if card.category == category
+        category_cards << card
+      end
+    end
+    return category_cards
```
epintozzi, line 20: Similar to the above, we don't need the `return` here.

spec/turn_spec.rb
```diff
@@ -0,0 +1,53 @@
+require 'rspec'
+require './lib/turn'
+require './lib/card'
+
+describe Turn do
+  it 'Card exists' do
```
epintozzi, line 6: It's a little redundant to test that a `Card` can exist from within the `turn_spec` since we've already tested that in the `card_spec`. However, it would be useful to test that a `Turn` has a card, so something like:

```ruby
  it 'Has a card' do
    card = Card.new("What is the capital of Oklahoma?", "Oklahoma City", :Geography)
    turn = Turn.new("Oklahoma City", card)

    expect(turn.card).to eq card
  end
  ```

File: lib/turn.rb [hunk 4]
```diff
@@ -0,0 +1,24 @@
+class Turn
```
epintozzi, line 1: As with the `Deck` class, we do not need to explicitly use `return` in the `correct?` and `feedback` methods.

lib/round.rb
```diff
@@ -0,0 +1,83 @@
...
+    @turns << turn
+    if turn.correct?
+      @number_correct += 1
+    end
+    return turn
+  end
+
+  def number_correct
+    return @number_correct
+  end
```
epintozzi, line 58: This method is the written out form of what `attr_reader :number_correct` is doing. You only need to have it here once so this method can be removed.

lib/round.rb
```diff
@@ -0,0 +1,83 @@
...
+
+  def number_correct_by_category(category)
+    category_correct = 0
+    @turns.each do |turn|
+      if category == turn.card.category
+          if turn.correct?
+            category_correct += 1
+          end
+        end
```
epintozzi, line 72: To make this a little cleaner, you could combine lines 68 & 69 so you only have one `if` statement:
```ruby
if category == turn.card.category && turn.correct?
  category_correct += 1
end
```

spec/round_spec.rb
```diff
@@ -0,0 +1,186 @@
...
+    card3 = Card.new("What do the letters GPU stand for in computing?", "Graphics Processing Unit", :Technology)
+    card4 = Card.new("What does NBA stand for in US Sports?", "National Basketball Association", :Sports)
+
+    cards = [card1, card2, card3, card4]
+
+    deck = Deck.new(cards)
+
+    round = Round.new(deck)
+
+    new_turn = round.take_turn("Oklahoma City")
```
epintozzi, line 96: If we aren't using a variable we can call the method without setting it equal to anything. Line 96 could be rewritten as: `round.take_turn("Oklahoma City")`

spec/round_spec.rb
```diff
@@ -0,0 +1,186 @@
+require 'rspec'
+require './lib/card'
+require './lib/deck'
+require './lib/turn'
+require './lib/round'
+
+# No tests included for start, end, or final_score methods due to dependence on input/output behavior
+
+describe Round do
```
epintozzi, line 9: I probably would've also tested the `final_score` method. It would take a little bit of setup, similar to the setup for testing the percentage calculations, but it would help verify the logic is working as intended, which is especially useful in case you ever decide to refactor something.

flashcard_runner.rb
```diff
@@ -0,0 +1,15 @@
...
+
+filename = "cards.txt"
+card_generator = CardGenerator.new(filename)
+card_generator.add_cards
+deck = Deck.new(card_generator.cards)
+round = Round.new(deck)
+
+round.start
+round.end
+round.final_score
```
epintozzi, line 15: 🔥  I love how simple your runner is

epintozzi (commented):
Ruby is really nice to us in that it will return the result of the last line of a method for us automatically so we do not need to explicitly call `return` in our methods. I identified this on a few of your methods but would be true for all of them.

Overall great work! I will follow up with scores in Slack by EOD Friday.

PR #389, opened 2022-05-15
John Santosuosso -Flashcards 5_15_22

JohnSantosuosso:
Thank you!


lib/deck.rb
```diff
@@ -0,0 +1,21 @@
...
+  def initialize(cards)
+  @cards = cards
+  end
+
+  def count
+    @cards.length
+  end
+
+  def cards_in_category(category)
+    @cards.select { |card| card.category == category }
```
abdulredd, line 16: Great use of the `select` enumerable here!

lib/turn.rb
```diff
@@ -0,0 +1,24 @@
...
+    @guess = guess
+    @card = card
+  end
+
+  def correct?
+    @guess.downcase.capitalize == card.answer.downcase.capitalize ? true : false
+  end
+
+  def feedback?
+    if correct? == true
```
abdulredd, line 16: Because a call to `correct?` returns a boolean you do not need to compare to a boolean to make sure it is `true`

These both are equivalent:

```ruby
if correct? == true
```

```ruby
if correct?
```

lib/flashcard_runner.rb
```diff
@@ -0,0 +1,33 @@
...
+    count += 1
+  end
+  puts "****** Game over! ******"
+  puts "You had #{round.number_correct} out of #{round.deck.count} for a total score of #{round.percent_correct}%"
+  puts "STEM - #{round.number_correct_by_category(:STEM)}% correct"
+  puts "Culture - #{round.number_correct_by_category(:Culture)}% correct"
+  puts "Geography - #{round.number_correct_by_category(:Geography)}% correct"
+end
+
+start(round)
```
abdulredd, line 33: The `start` method is actually supposed to be an instance method on your `Round` class and not a global method that you pass `round` into. This `flashcard_runner.rb` file should only have the setup code the call to `round.start`

```ruby
#[YOUR REQUIRE STATEMENTS]

cards = CardGenerator.new("cards.txt").cards
deck = Deck.new(cards)
round = Round.new(deck)

round.start
```

lib/flashcard_runner.rb
```diff
@@ -0,0 +1,33 @@
...
+card_5 = Card.new("What river divides Washington DC from Virginia?", "Potomac", :Geography)
+cards = [card_1, card_2, card_3, card_4, card_5]
+deck = Deck.new(cards)
+round = Round.new(deck)
+
+def start(round)
+  count = 1
+  puts "Welcome! You're playing with #{round.deck.count} cards."
+  puts '-------------------------------------------------'
+  while count < round.deck.count
```
abdulredd, line 19: Because this checks if `count < round.deck.count` it always ends the game one card short of the final card. You did correct it in your `flashcard_runner1.rb` file so I believe you can identify the issue.

lib/flashcard_runner1.rb
```diff
@@ -0,0 +1,32 @@
...
+require_relative 'card_generator'
+
+filename = "cards.txt"
+cards = CardGenerator.new(filename).cards
+deck = Deck.new(cards)
+round = Round.new(deck)
+
+def start(round)
+  count = 1
+  puts "Welcome! You're playing with #{round.deck.count} cards."
```
abdulredd, line 14: Great use of interpolation here, that makes your game dynamic. No matter how many cards are in the deck, the correct number always prints.

lib/flashcard_runner1.rb
```diff
@@ -0,0 +1,32 @@
...
+  while count <= round.deck.count
+    puts "\nThis is card number #{count}"
+    puts round.deck.cards.first.question
+    guess = gets.chomp
+    puts round.take_turn(guess).feedback?
+    count += 1
+  end
+  puts "****** Game over! ******"
+  puts "You had #{round.number_correct} out of #{round.deck.count} for a total score of #{round.percent_correct}%"
+  puts "STEM - #{round.number_correct_by_category(:STEM)}% correct"
```
abdulredd, line 25: Hard-coding the categories here is something you want to avoid. This only works if the cards in your deck are of STEM, Culture, or Geography category types but doesn't count for potential cards of other categories making it less dynamic.

PR #390, opened 2022-05-15
Parker Thomson

pdthomson:


lib/card.rb
```diff
@@ -0,0 +1,12 @@
...
+  attr_reader :question,
+              :answer,
+              :category
+
+  def initialize(question, answer, category)
+    @question = question
+    @answer = answer
+    @category = category
+  end
+end
```

lib/deck.rb
```diff
@@ -0,0 +1,16 @@
...
+  end
+
+  def count
+    @cards.length
+  end
+
+  def cards_in_category(category)
+    @cards.select {|card| category == card.category}
+  end
+end
```
megstang, line 16: 👍

lib/round.rb
```diff
@@ -0,0 +1,61 @@
...
+  def number_correct
+    @correct_turns.count
+  end
+
+  def number_correct_by_category(category)
+    correct = @correct_turns.select do |turn|
+      category == turn.card.category
+    end
+    correct.length
+  end
```
megstang, line 36: nice use of select here!

lib/round.rb
```diff
@@ -0,0 +1,61 @@
+class Round
+
+  attr_reader :deck,
+              :turns,
+              :correct_turns
```
megstang, line 5: I don't think you need to call a`.correct_Turns` on a round object anyhwere outside of this class. So, for that reason, I'd remove this attr_reader, and just use your instance variable you declared on line 10 when you need to reference that value throughout this class.

lib/turn.rb
```diff
@@ -0,0 +1,26 @@
...
+    @card = card
+  end
+
+  def correct?
+    if @guess == @card.answer
+      true
+    else
+      false
+    end
+  end
```
megstang, line 17: You could slim this up quite a bit by saying 

```
def correct? 
 @ guess == @card.answer
end 
```

We could do this becuase `==` will automatically return true or false for us when it's evaluated.

lib/turn.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def feedback
+    if correct? == true
```
megstang, line 20: similarly here, since `correct?` returns true or false automatically, we don't need to have this boolean operator. So, this line could just read `if correct?`

spec/turn_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  it "can check if the guess is correct" do
+  expect(@turn.correct?).to eq(true)
+  end
+
+  it "can tell you if your guess was correct or not" do
+  expect(@turn.feedback).to eq("Correct!")
+  end
+end
```
megstang, line 26: Make sure you're testing the case for when a guess is incorrect.

megstang (commented):
I left some comments for you to look over. Please let me know if you have any questions. I'll follow up in a slack DM later today with your scores.

PR #391, opened 2022-05-15
Shirley DeCesari 

sdecesari:


lib/turn.rb
```diff
@@ -0,0 +1,20 @@
...
+    @guess = guess
+    @card = card
+  end
+
+  def correct?
+    @guess.capitalize == @card.answer.capitalize
+  end
+
+  def feedback
+    if correct? == true
```
epintozzi, line 14: This line can also be written as `if correct?` because the `if` is looking for a true or false result and `correct?` on its own is true (or false).

lib/round.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  def take_turn(guess)
+    @turns << Turn.new(guess, current_card)
+    @deck.cards.rotate!(1)
+    @turns.last
+  end
+
+  def number_correct
+    @turns.count{ |turn| turn.correct?}
```
epintozzi, line 20: This is great! Another way you can write this is:
`@turns.count(&:correct?)`

Neither is better than the other, but this class is already very well written and refactored, I thought you might like to see some other ruby tricks.

epintozzi (commented):
Overall this looks really nice. You have great ruby style/syntax/formatting. The code is clean and easy to read. You aren't using instance variables and attr_readers where you don't need them which is also great. This is also well tested. Good work! Official scores will be sent out via Slack by Friday end of day.

PR #392, opened 2022-05-15
Jared Hardinger [1]

jaredhardinger:
Complete first three iterations of project

### DIFF
--- a/lib/card.rb
+++ b/lib/card.rb
@@ -1,0 +1,11 @@
+class Card
+  attr_reader :question, :answer, :category
+  def initialize(question, answer, category)
+    @question = question
+    @answer = answer
+    @category = category
+  end
+
+
+
+end
--- a/lib/deck.rb
+++ b/lib/deck.rb
@@ -1,6 +1,19 @@
 class Deck
+  attr_reader :cards, :count
+  def initialize (cards)
+    @cards = cards
+    @count = cards.count
+  end
 
+  def cards_in_category(category)
+    same_cat = []
+    @cards.each do |card|
+      if card.category == category
+        same_cat << card
+      end
+    end
+    same_cat
 
-
+  end
 
 end
--- a/lib/deck.rb
+++ b/lib/deck.rb
@@ -1,0 +1,6 @@
+class Deck
+
+
+
+
+end
--- a/lib/deck.rb
+++ b/lib/deck.rb
@@ -6,13 +6,13 @@ class Deck
   end
 
   def cards_in_category(category)
-    same_cat = []
+    same_category = []
     @cards.each do |card|
       if card.category == category
-        same_cat << card
+        same_category << card
       end
     end
-    same_cat
+    same_category
 
   end
 
--- a/lib/round.rb
+++ b/lib/round.rb
@@ -46,10 +46,18 @@ class Round
       if turn.card.category == category
         number_in_category += 1
       end
-    end 
-    number_correct_by_category(category) * 100 / number_in_category
+    end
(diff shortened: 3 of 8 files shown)


lib/card.rb
```diff
@@ -0,0 +1,11 @@
+class Card
+  attr_reader :question, :answer, :category
+  def initialize(question, answer, category)
+    @question = question
+    @answer = answer
+    @category = category
+  end
+
+
+
```
megstang, line 10: get rid of this extra space

lib/deck.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  def cards_in_category(category)
+    same_category = []
+    @cards.each do |card|
+      if card.category == category
+        same_category << card
+      end
+    end
+    same_category
+
```
megstang, line 16: remove extra space

lib/deck.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def cards_in_category(category)
+    same_category = []
+    @cards.each do |card|
+      if card.category == category
+        same_category << card
+      end
+    end
+    same_category
```
megstang, line 15: great job here! We haven't talked much about other enumerables yet, but this could be a great method for using `find_all`.

lib/round.rb
```diff
@@ -0,0 +1,79 @@
+class Round
+  attr_reader :deck, :turns, :deck_index
```
megstang, line 2: you never need to call `.deck_index` on an instance of the round class in other classes, so `:deck_index` does not need to be an  attr_reader.

megstang: it's kind of like how you don't have an attr_Reader for number_correct -- you just need that value within the class, but nowhere else.

lib/round.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  def current_card
+    deck.cards[@deck_index]
+  end
+
+  def take_turn(guess)
+    new_turn = Turn.new(guess, current_card)
+    @turns << new_turn
+    if new_turn.correct?
+
```
megstang, line 18: remove extra space

lib/round.rb
```diff
@@ -0,0 +1,79 @@
...
+  end
+
+  def percent_correct_by_category(category)
+    number_in_category = 0
+    turns.each do |turn|
+      if turn.card.category == category
+        number_in_category += 1
+      end
+    end
+    number_correct_by_category(category) * 100.0 / number_in_category
```
megstang, line 50: nice job calling to an already written method to get this logic done.

spec/deck_spec.rb
```diff
@@ -0,0 +1,30 @@
+require './lib/card'
+require './lib/deck'
+
+RSpec.describe Deck do
+  card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+  card_2 = Card.new("The Viking spacecraft sent back to Earth photographs and reports about the surface of which planet?", "Mars", :STEM)
+  card_3 = Card.new("Describe in words the exact direction that is 697.5° clockwise from due north?", "North north west", :STEM)
+  deck = Deck.new([card_1, card_2, card_3])
```
megstang, line 8: I would recommend not declaring variables outside of your it blocks. This makes them class variables, which means, as you go about your tests, running methods or passing those certain variables around, it's possible you're changing the attributes associated with each variable. For that reason, I'd define the variables in each it block, OR, look into before(:each) if you're really interested in slimming up your files.

megstang (commented):
I left some comments for you to look over. Please let me know if you have any questions. I'll follow up sometime tomorrow with your scores.

PR #393, opened 2022-05-15
Ethan Nguyen

Ethan-t-n:


flashcard_runner.rb
```diff
@@ -0,0 +1,45 @@
...
+cards = CardGenerator.new("cards.txt").cards
+
+
+deck = Deck.new(cards)
+
+round = Round.new(deck)
+
+round.start
+
+i = 1
```
abdulredd, line 23: Lines **23 - 45** should not be in the `flashcard_runner.rb` file. This code should be inside of the `start` method of your `Round` class.

The runner file should only contain setup code and the entry point to fire off your application.

```ruby
cards = CardGenerator.new("cards.txt").cards
deck = Deck.new(cards)
round = Round.new(deck)

round.start
```

The code that you have between those lines works, its just unfortunately in the wrong place!

lib/deck.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  def count
+    @cards.length
+  end
+
+
+
+
+  def cards_in_category(category)
+    cards.select do |card|
```
abdulredd, line 20: Great use of the `select` enumerable here!

lib/turn.rb
```diff
@@ -0,0 +1,20 @@
...
+    @guess = guess
+    @card = card
+  end
+
+  def correct?
+    return guess == card.answer
+  end
+
+  def feedback
+    return "Correct!" if correct? == true
```
abdulredd, line 14: You don't need to check the return value of the call to `correct?` because it's either going to return `true` or `false` so you don't need to verify it.

These both mean the same thing:

```ruby
if correct? == true
```

```ruby
if correct?
```

There really isn't a need for 2 returns here and this method could have been written like so:

```ruby
def feedback
  if correct?
    true
  else
    false
  end
end
```

```ruby
def feedback
  if correct?
    return true
  end
  false
end
```

lib/round.rb
```diff
@@ -0,0 +1,85 @@
...
+       number_correct += 1
+     end
+   end
+   percent_correct = (number_correct.to_f / @turns.count.to_f) * 100
+   return percent_correct
+
+  end
+
+  def start
+   start_message = "Welcome! You're playing with #{deck.cards.count} cards."
```
abdulredd, line 73: Great use of interpolation here to make it actually use the number of cards you have in your deck. I added a few extra cards of my own and the number changed accordingly. Good job!

PR #395, opened 2022-05-15
Justin Ramirez

jusrez:


lib/card.rb
```diff
@@ -0,0 +1,12 @@
...
+							:answer,
+							:category
+	def initialize (question, answer, category)
+		@question = question
+		@answer = answer.downcase
+		@category = category
+
+	end
+
+end
```
megstang, line 12: This looks good. I like that you're downcasing your answer for edge cases. Make sure to check your spacing before pushing up to github, as syntax is a bit off.

lib/deck.rb
```diff
@@ -0,0 +1,21 @@
...
+	def cards_in_category(category)
+		cards_in_category = []
+		@cards.each do |card|
+			if card.category == category
+				cards_in_category << card
+			end
+		end
+		cards_in_category
+	end
+end
```
megstang, line 21: 👍  nice work

lib/round.rb
```diff
@@ -0,0 +1,89 @@
...
+	end
+
+	def percent_correct_by_category(category)
+		total_in_category = 0
+		@turns.each do |turn|
+			if turn.card.category == category
+				total_in_category += 1
+			end
+		end
+		(number_correct_by_category(category).to_f / total_in_category) * 100
```
megstang, line 53: I like that you are referencing your previous method, number_correct_by_category to help you out here.

lib/turn.rb
```diff
@@ -0,0 +1,21 @@
+class Turn
+	attr_reader :guess, :card
+
+	def initialize(guess, card)
+		@guess = guess
+		@card = card
+	end
+
+	def correct?
+		@guess == card.answer
```
megstang, line 10: instead of calling to `card`, it's more conventional to call to `@card` in this scenario. Since `card` is referencing your attr_reader, which returns `@card`, you might as well call directly to `@card` and cut out the middle man.

lib/turn.rb
```diff
@@ -0,0 +1,21 @@
...
+	end
+
+	def feedback
+		if correct? == true
+		"Correct!"
+		elsif correct? == false
+		"Incorrect."
+		end
+	end
```
megstang, line 19: since `correct?` evaluates to true or false automatically, you could slim up this method a bit by doing something like: 

```
	def feedback
		if correct?
		"Correct!"
		else
		"Incorrect."
		end
	end
```

spec/round_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+		card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+		card_2 = Card.new("The Viking spacecraft sent back to Earth photographs and reports about the surface of which planet?", "Mars", :STEM)
+		card_3 = Card.new("Describe in words the exact direction that is 697.5° clockwise from due north?", "North north west", :STEM)
+		deck = Deck.new([card_1, card_2, card_3])
+		round = Round.new(deck)
+		new_turn = round.take_turn("Juneau")
+		round.take_turn("Venus")
+
+		expect(round.percent_correct).to eq(50.0)
+	  expect(round.percent_correct_by_category(:Geography)).to eq(100.0)
```
megstang, line 67: might be nice to have just a few more assertions for testing this method.

megstang: for example, if this method always returned 100, your tests wouldn't know that, the more instance you test, the more you can be sure that your code is working for all cases.

megstang (commented):
Nice work! I left a few comments for you to look over. Please let me know if you have any questions. I'll follow up in slack with your scores.

PR #396, opened 2022-05-15
Andre Pedro

tenworms:


lib/deck.rb
```diff
@@ -0,0 +1,25 @@
...
+class Deck
+
+  attr_reader :cards
+
+  def initialize(cards)
+    @cards = cards
+  end
+
+  def count
+    cards.count
```
cjsim89, line 11: Be careful when accessing instance variables, you want to always reference instance variables by `@`, even though Ruby lets us do it without it. This consistency will make more complex code easier to read later on!

lib/round.rb
```diff
@@ -0,0 +1,86 @@
...
+
+
+
+  def take_turn(new_guess)
+     turn = Turn.new(new_guess, current_card)
+     @turns << turn
+     if new_guess == current_card.answer
+       @number_correct += 1
+     end
+     deck.cards.rotate!
```
cjsim89, line 23: Nice job using this built-in method. Be careful when using `!` methods, as they can inadvertently change the base object you're working on, and may affect things later on down the road. But for now, it works great.

lib/round.rb
```diff
@@ -0,0 +1,86 @@
...
+       if turn.card.category == category && turn.correct?
+         number_correct_by_category += 1
+       end
+     end
+     return number_correct_by_category
+   end
+
+   def percent_correct
+     percent = (number_correct.to_f / turns.count) *100
+     return percent
```
cjsim89, line 40: I notice you're using a lot of explicit `return` statements, which is fine - try to get used to using the "implicit" return as well, it tends to make code a lot cleaner overall.

lib/round.rb
```diff
@@ -0,0 +1,86 @@
...
+       end
+     end
+     (number_correct_by_category(category).to_f / total_in_category) * 100
+   end
+
+
+   def start
+      i = 0
+      while i < deck.count
+        i += 1
```
cjsim89, line 59: Using a `while` loop + an incrementor like `i` is cool syntax, you might learn some other ways Ruby has this built-in with different enumerables later on!

lib/round.rb
```diff
@@ -0,0 +1,86 @@
...
+        puts "Question #{i} of #{deck.count}:  #{current_card.question}"
+        puts take_turn(gets.chomp).feedback
+      end
+   end
+
+
+   # p "You had #{round.number_correct} correct guesses out of #{deck.count} for a total score of #{round.percent_correct.to_i}%"
```
cjsim89, line 66: It's good practice to remove comments from your code unless they help with readability. If this was stuff you were trying out, you can always put them into other branches than `main`, for reference later on.

lib/turn.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def initialize(guess, card)
+    #@turn = []
+    @card = card
+    #@turn_counter = []
+    @guess = guess
+  end
+
+  def correct?
+     @guess == card.answer
```
cjsim89, line 13: Again, be careful about referencing instance variables (like `@card`) without the `@` sign.

spec/round_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+    cards = [card_1, card_2, card_3]
+    deck = Deck.new(cards)
+    round = Round.new(deck)
+
+
+    expect(round.deck).to eq(deck)
+    expect(round.turns).to eq([])
+  end
+
+  it 'current card' do
```
cjsim89, line 35: This is fine, but it can be helpful to be more explicit in your test names about what exactly you're testing. For example, this test might be called `it 'has the first card in the deck as current_card' do`.

spec/round_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+    deck = Deck.new(cards)
+    round = Round.new(deck)
+
+    expect(round.current_card).to eq(card_1)
+
+  end
+
+
+
+  it 'takes turn' do
```
cjsim89, line 49: Same here - a better test name might be `it "has a correct turn guess" do`.

cjsim89 (commented):
Hi @tenworms - nice job on your first project! I left some feedback in the form of comments on your PR, let me know if you have any questions 👍

PR #397, opened 2022-05-15
Mary Turpin

MaryT573:


lib/card.rb
```diff
@@ -0,0 +1,10 @@
+class Card
+
+  attr_reader :category, :question, :answer
+
+  def initialize(question, answer, category)
+    @question = question
+    @answer = answer
+    @category = category
+  end
+end
```

lib/deck.rb
```diff
@@ -0,0 +1,23 @@
+class Deck
+
+  attr_reader :cards, :count, :cards_in_category
```
megstang, line 3: since you have a method for cards_in_category, you don't need to have this attr_reader. We only create attr_readers when we have an attribute that we want to be able to read outside of the class. Since you have handmade a method for cards_in_category, this attr_reader is moot, in fact, scope-wise, ruby will look at your handmade method first before it looks at this attr_Reader, so it's never actually hit.

lib/deck.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def initialize(cards)
+    @cards = cards
+  end
+
+  def count
+    cards.count
```
megstang, line 10: Instead of calling to `cards`, call to `@cards`. When you call to cards, it calls to the attr_Reader, which returns the instance variable. Since you can call directly to the instance variable, do that, and cut out the middle man.

lib/deck.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def cards_in_category(category)
+    card_category = []
+      cards.each do |card|
+        if category == card.category
+          card_category << card
+        end
+      end
+    card_category
+  end
```
megstang, line 21: 👍 nice!

lib/round.rb
```diff
@@ -0,0 +1,67 @@
...
+    new_turn = Turn.new(new_guess, current_card)
+    #require 'pry'; binding.pry
+    @turns << new_turn
+    if new_guess == current_card.answer
+      @number_correct += 1
+    end
+    deck.cards.rotate!
+    #might need to edit
+    return new_turn
+  end
```
megstang, line 28: make sure that when you push up your code, that you clean it up. Its totally fine to leave comments that describe what your code is doing, but any old code that you commented out, or prys, remove those lines before pushing up.

lib/round.rb
```diff
@@ -0,0 +1,67 @@
...
+      if turn.card.category == category && turn.correct?
+        number_correct_by_category += 1
+      end
+    end
+    return number_correct_by_category
+  end
+
+ def percent_correct
+    percent = (number_correct.to_f/turns.count) * 100
+    return percent
```
megstang, line 43: you don't necessarily need line 43. This value that gets calculated on 42 will automatically get returned from this method if that's the last line of your method.

megstang (commented):
Nice work! I left some comments for you to look over. Please let me know if you have any questions. I'll follow up in slack with your scores later today.

PR #398, opened 2022-05-15
Ben Silverstein

bensjsilverstein:


lib/deck.rb
```diff
@@ -0,0 +1,24 @@
+class Deck
+
+  attr_reader :cards
+  def initialize(cards)
+     @cards = cards
+
+  end
+
+  def count
+    cards.count
```
cjsim89, line 10: Be careful when referencing instance variables that you always include the `@` - even though Ruby lets it work without it, it can be confusing in larger, more complex applications.

lib/deck.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def cards_in_category(category)
+    test = []
+    @cards.each do |card|
+      # test << category == card.category
+      if category == card.category
+        test << card
+      end
+    end
+    return test
```
cjsim89, line 21: I see you used an accumulator here - great - try to be more descriptive with your variable names. What does `test` tell me about the data contained within? (Answer: not much)

lib/round.rb
```diff
@@ -0,0 +1,104 @@
...
+    @deck = deck
+    @turns = []
+    @number_correct = 0
+    @turn_count = 0
+
+
+  end
+
+  def current_card
+    deck.cards[0]
```
cjsim89, line 15: Again, be careful about referencing instance variables without the `@`.

lib/round.rb
```diff
@@ -0,0 +1,104 @@
...
+  end
+
+  def take_turn(guess)
+
+    turn = Turn.new(guess, current_card)
+    turns << turn
```
cjsim89, line 21: Instance variables should have an `@` for consistency.

lib/round.rb
```diff
@@ -0,0 +1,104 @@
...
+    if turn.correct?
+      @number_correct += 1
+    end
+    deck.cards.rotate!
+    return turn
+    # turn
```
cjsim89, line 27: Looks like you went with an explicit return here instead of implicit - it can be weird not using the `return` keyword but it is totally fine to do so in Ruby! Try to get used to the joys of "letting go of the handlebars" when you can ;)

lib/round.rb
```diff
@@ -0,0 +1,104 @@
...
+      i += 1
+      end
+    end
+    return (number_correct_by_category(category).to_f / i.to_f)*100.0
+  end
+
+  def start
+    x = 0
+    while x < deck.count
+      x += 1
```
cjsim89, line 69: Nice use of `while` here! There might be some enumerables you learn in the future that can make this method a little easier to read.

lib/turn.rb
```diff
@@ -0,0 +1,22 @@
...
+  attr_reader :guess , :card
+
+  def initialize(guess, card)
+    @guess = guess
+    @card = card
+
+  end
+
+  def correct?
+    guess == card.answer
```
cjsim89, line 11: n+1 to the non-use of `@` when referencing instance variables. It might not seem like much now or in this example, but it will make a bigger difference when we start throwing around different types of scope.

cjsim89 (commented):
Nice job on your project @bensjsilverstein ! I left some comments on your PR, let me know if you have any questions.

PR #399, opened 2022-05-15
Gwendolyn Ruiz

gwen-marina:


flashcard_runner.rb
```diff
@@ -0,0 +1,21 @@
...
+card_6 = Card.new("Which is the only edible food that never goes bad?", "Honey", :Food)
+card_7 = Card.new("Au is the symbol for which element?", "Gold", :STEM)
+card_8 = Card.new("What is the top selling video game of all time?", "Minecraft", :Entertainment)
+card_9 = Card.new("What was the first soft drink in space?", "Coca Cola", :Food)
+card_10 = Card.new("What is the 5th planet from the sun?", "Jupiter", :STEM)
+deck = Deck.new([card_1, card_2, card_3, card_4, card_5, card_6, card_7, card_8, card_9, card_10])
+round = Round.new(deck)
+
+round.start
+round.game_end_stats
```
abdulredd, line 21: I would have liked to see this method `game_end_stats` called from within the `start` method. `game_end_stats` and `start` are highly dependent on each other, if `game_end_stats` were to happen to be called before `start`, the application would break. And conversely if `game_end_stats` were not called after the `start` method, the game would just end and no stats would be printed.

lib/turn.rb
```diff
@@ -0,0 +1,25 @@
...
+  def correct?
+    if @guess == @card.answer
+      true
+    else
+      false
+    end
+  end
+
+  def feedback
+    if correct? == true
```
abdulredd, line 19: Because a call to `correct?` already returns a boolean, you do not need to necessarily compare it to the value `true`. Omitting that part also works.

```ruby
def feedback
  if correct?
    "Correct!"
  else
    "Incorrect."
  end
end
```

lib/round.rb
```diff
@@ -0,0 +1,79 @@
...
+    @turns.each do |turn|
+      if turn.card.category == category
+        turns_by_category << turn
+      end
+    end
+    number_correct_by_category(category).to_f / turns_by_category.count * 100
+  end
+
+  def start
+    puts "Welcome! You're playing with #{deck.count} cards."
```
abdulredd, line 49: Great use of interpolation here and making the application dynamic!

lib/round.rb
```diff
@@ -0,0 +1,79 @@
...
+    turns = 0
+    until turns == deck.count do
+      turns += 1
+      puts "This is card number #{turns} out of #{deck.count}."
+      puts current_card.question
+      user_input = gets.chomp
+      turn = take_turn(user_input)
+      puts turn.feedback
+    end
+      puts "****** Game over! ******"
```
abdulredd, line 60: Like I mentioned earlier I would have put the call to `game_end_stats` at the end of the `start` method since they are highly dependent on each other.

```ruby
  game_end_stats
end
```

This way you don't have to put it in the `flashcard_runner.rb` file

lib/round.rb
```diff
@@ -0,0 +1,79 @@
...
+  end
+
+  def turns_by_category
+    categories = []
+    @deck.cards.count do |card|
```
abdulredd, line 65: This isn't breaking but I believe this is a syntax error and Ruby is probably doing some magic behind the scenes and helping you out a bit here. What I think you wanted here was:

```ruby
@deck.cards.each do |card|
```

also on line **66** you have a `card.category` statement that is not being used and can probably be removed without affecting anything.

PR #401, opened 2022-05-15
Gauri Joshi

gaurijo:


flashcard_runner.rb
```diff
@@ -0,0 +1,247 @@
+require './lib/round.rb'
+#create some Cards
+#put those card into a Deck
+#create a Round using the Deck you created
+#START the round using a new method called start
```
cjsim89, line 5: This is a really good practice - take what you read from the specifications/interaction patterns, and put them into your own words in comments! Keep doing this, I think it can be really helpful.

flashcard_runner.rb
```diff
@@ -0,0 +1,247 @@
...
+class Card
```
cjsim89, line 6: The runner file should not contain any other code than what is needed to run the program. Your `Card` class already exists at `./lib/card`, so we just need to `require` this and any other classes your program needs to run at the top of this runner file. 

That would look like this: 
```
require './lib/card'
require './lib/deck'
require './lib/turns'
```
I see you already have the `Round` class required at the top, that's the right idea. Copying in a class over here defeats the purpose of your tests, because now we have 2 classes doing the same things, at 2 different locations. If this runner file runs, it will be using the `Card` class that's _here_, and not the one in your `lib` folder. This can cause all sorts of problems. Next time, just `require` your classes wherever you need them.

flashcard_runner.rb
```diff
@@ -0,0 +1,247 @@
...
+card_1 = Card.new("What is the fastest animal?", "Cheetah", :Animal)
+card_2 = Card.new("What is the color of the ocean?", "Blue", :Geography)
+card_3 = Card.new("What is the coolest city on earth?", "NYC", :Geography)
+card_4 = Card.new("What cardboard cutout lives at Turing?", "Justin Bieber", :Pop)
+
+cards = [card_1, card_2, card_3, card_4]
+
+# p cards
+
+class Deck
```
cjsim89, line 24: Same as above - the `Deck` class should be `require`d at the top of your runner file.

flashcard_runner.rb
```diff
@@ -0,0 +1,247 @@
...
+  def initialize(cards)
+    @cards = cards
+    @category = category
+  end
+end
+
+deck = Deck.new(cards)
+# p deck
+
+class Turns
```
cjsim89, line 35: Same - `require './lib/turns`

flashcard_runner.rb
```diff
@@ -0,0 +1,247 @@
...
+  end
+
+  def correct?
+    if @guess == @answer
+      return true
+    end
+  end
+
+  def feedback
+    if @guess == @answer
```
cjsim89, line 50: This is repeated code - we're already doing `if @guess == @answer` up on line 44; maybe we call `correct?` here instead...?

flashcard_runner.rb
```diff
@@ -0,0 +1,247 @@
...
+
+# round.take_turn("Cheetah")
+
+puts turns.feedback
+# round.add_1_to_correct
+
+puts "This is card 2 out of 4."
+puts cards[1].question
+
+@guess = gets.chomp
```
cjsim89, line 87: Since we're in the context of the runner file, and not the `Turns` or `Round` classes, where does `@guess` get read...? 

Asking because, when I run this runner file, it never says my guesses are correct, even though I am copy-pasting the answers from your cards. The reason is that we're setting `@guess` here (and on the other turns in this runner file as well), but the value is never actually being read inside the `Turns` class, because `@guess` has context only there, and not here. We need to __send in__ the value of the user's guess (the return value of `gets.chomp`) to the Turn class a different way.

flashcard_runner.rb
```diff
@@ -0,0 +1,247 @@
...
+
+turns_2 = Turns.new("Blue", cards[1])
+
+round.add_1_to_turn
+turns_2.correct?
```
cjsim89, line 92: The `.correct?` method returns `true` or `nil`. Since it returns something here, it should be set to a variable of some kind, or evaluated in a conditional statement (`if turns_2.correct? ... { do some code } ... `). Right now, it's just sort of hanging out.

flashcard_runner.rb
```diff
@@ -0,0 +1,247 @@
...
+# round.add_1_to_correct
+
+
+print "***********GAME OVER***************"
+
+puts
+
+
+
+puts "You had 4 correct guesses out of 4 for a score of  "
```
cjsim89, line 136: I think you may have been in the middle of coding this out - we would want the #s in this line to be variables. Something like... 
 `"You had #{num_correct} correct guesses out of #{total_number_of_cards} for a score of" ` would be how I would start this out. Then, those variables would need to come from one of your classes.

flashcard_runner.rb
```diff
@@ -0,0 +1,247 @@
...
+p round.percent_correct(4,4)
```
cjsim89, line 137: Same here - you're sending in 4 and 4 for `x` and `y`, but these should be variables based on how the user performed.

flashcard_runner.rb
```diff
@@ -0,0 +1,247 @@
...
+
+p "Animal - #{round.percent_correct_by_category(:Animal)} % correct"
```
cjsim89, line 139: This line is correct, but the logic inside the `percent_correct_by_category` is not, so this does not print the right thing.

lib/deck.rb
```diff
@@ -0,0 +1,25 @@
...
+    @category = category
+
+  end
+
+  def count
+    return @cards.count
+  end
+
+  def cards_in_category(cat)
+    if cat == :STEM
```
cjsim89, line 14: I can see what you're going for here, but `:STEM` is a hard-coded value. In your runner file, you use different categories altogether, so this code will never execute correctly when we run the runner file. These do make your tests pass, but we should not use hard-coded values like this in our classes, when we can avoid it.

lib/deck.rb
```diff
@@ -0,0 +1,25 @@
...
+      return cards[1,2]
```
cjsim89, line 15: This method could look something like: 
```
cards_to_return = [] # an accumulator would work here 
@cards.each do |card|
   if card.category == cat #where cat is the argument that we pass in on line 13
      cards_to_return << card
   end
end
return cards_to_return
```

cjsim89: Otherwise, selecting `cards[1,2]` would always return the 2nd and 3rd cards in our array. But, what if the 3rd card that we use in a game is NOT in the category `:STEM`?

lib/turns.rb
```diff
@@ -0,0 +1,35 @@
...
+  def initialize(guess, card)
+    @guess = guess
+    @card = card
+    @answer = answer
+  end
+
+  def correct?
+    if @guess == @answer
+    #"Juneau"
+    return true
```
cjsim89, line 13: This works, but there is no condition for if guess does not equal answer. In this case, it would return `nil`, but for cases like this, it is usually better to return `false` inside of an `else`, instead of `nil`. I find this easier to understand when debugging; usually if I encounter a `nil`, it makes me more confused than a `false`. 

Here's what this could look like: 
```
if @guess == answer 
   return true
else 
   return false
end
```

cjsim89: To be clear - yours works, but including a `return false` inside the `else` makes the code a little easier to read, and easier to test against.

lib/turns.rb
```diff
@@ -0,0 +1,35 @@
...
+      return "Incorrect!"
+    end
+
+  end
+
+end
+
+turns = Turns.new("Juneau", @card)
+turns.correct?
+turns.feedback
```
cjsim89, line 30: Lines 28 - 30 is outside of the class, and so should not be included in the class file. This will probably make `require`-ing this class in other files problematic, it might behave in a strange way.

spec/round_spec.rb
```diff
@@ -0,0 +1,171 @@
...
+
+RSpec.describe Round do
+  it 'exists' do
+    card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    card_2 = Card.new("The Viking spacecraft sent back to Earth photographs and reports about the surface of which planet?", "Mars", :STEM)
+    card_3 = Card.new("Describe in words the exact direction that is 697.5° clockwise from due north?", "North north west", :STEM)
+
+    cards = [card_1, card_2, card_3]
+    deck = Deck.new(cards)
+    turns = Turns.new("Juneau", @card)
```
cjsim89, line 14: Where does `@card` come from...? Remember, instance variables are based on the **scope** of a class or instance; another word for scope is "context". `@card` has no context in this file, because it belongs in the Deck class. So, instead of passing in `@card` in your tests, you should pass in `card_1` or one of the cards you made on lines 8-10.

spec/round_spec.rb
```diff
@@ -0,0 +1,171 @@
...
+  end
+
+  it 'returns a deck' do
+    card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    card_2 = Card.new("The Viking spacecraft sent back to Earth photographs and reports about the surface of which planet?", "Mars", :STEM)
+    card_3 = Card.new("Describe in words the exact direction that is 697.5° clockwise from due north?", "North north west", :STEM)
+
+    cards = [card_1, card_2, card_3]
+    deck = Deck.new(cards)
+    turns = Turns.new("Juneau", @card)
```
cjsim89, line 28: Same as above - use `card_1` since this instance variable has no context here.

spec/round_spec.rb
```diff
@@ -0,0 +1,171 @@
...
+  end
+
+  it 'has turns' do
+    card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    card_2 = Card.new("The Viking spacecraft sent back to Earth photographs and reports about the surface of which planet?", "Mars", :STEM)
+    card_3 = Card.new("Describe in words the exact direction that is 697.5° clockwise from due north?", "North north west", :STEM)
+
+    cards = [card_1, card_2, card_3]
+    deck = Deck.new(cards)
+    turns = Turns.new("Juneau", @card)
```
cjsim89, line 42: Same as above - use card_1 since this instance variable has no context here.

spec/round_spec.rb
```diff
@@ -0,0 +1,171 @@
...
+  end
+
+  it 'has a current card' do
+    card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    card_2 = Card.new("The Viking spacecraft sent back to Earth photographs and reports about the surface of which planet?", "Mars", :STEM)
+    card_3 = Card.new("Describe in words the exact direction that is 697.5° clockwise from due north?", "North north west", :STEM)
+
+    cards = [card_1, card_2, card_3]
+    deck = Deck.new(cards)
+    turns = Turns.new("Juneau", @card)
```
cjsim89, line 56: Same as above - use card_1 since this instance variable has no context here.

spec/turns_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  it 'is correct' do
+    turns = Turns.new("Juneau", @card)
+    expect(turns.correct?).to eq(true)
+
+  end
+
+  it 'has feedback' do
+    turns = Turns.new("Juneau", @card)
```
cjsim89, line 33: Same as above - use `card` since this instance variable has no context here.

I think you meant to use the `card` listed on line 7 of this file - it needs to be included inside of each `it` block for this to work. Right now, this test is failing, along with the one on line 28, for the same reason.

spec/round_spec.rb
```diff
@@ -0,0 +1,171 @@
...
+    round = Round.new(deck)
+
+    new_turn_1 = round.take_turn("Juneau")
+    new_turn_2 = round.take_turn("Mars")
+
+    round = Round.new(deck)
+
+    expect(new_turn_1.correct?).to be(true)
```
cjsim89, line 94: This is actually returning `nil`, so this test is failing. 

It is returning `nil` because there is no `current_card` in the Turns class at this moment - check line 29 in your `round.rb` file to be sure.

PR #404, opened 2022-05-15
Drew MacNicholas

dmacnicholas:


lib/round.rb
```diff
@@ -0,0 +1,82 @@
...
+        end
+    end
+    category_correct.count
+  end
+
+  def percent_correct
+    (@correct_cards.count.to_f / @turns.count.to_f) * 100
+  end
+
+  def percent_correct_by_category(category)
```
cjsim89, line 44: Looks like you got a start on iteration 4 - might be good to come back to this someday and finish it out, for a good portfolio piece during the job hunt!

lib/round.rb
```diff
@@ -0,0 +1,82 @@
...
+  def take_turn(guess)
+    turn = Turn.new(guess, current_card)
+    @turns << turn
+    if turn.correct?
+      @number_correct += 1
+      @correct_cards << turn.card
+    end
+    turn
+  end
+  # Do not need this method since it is also defined above in reader
```
cjsim89, line 25: Bingo - leaving it in causes the program to run a little slower (infinitesimally so, but still). Code would be a little more readable by depending only on the attr_reader.

lib/round.rb
```diff
@@ -0,0 +1,82 @@
...
+
+  def initialize(deck)
+    @deck = deck
+    @turns = []
+    @number_correct = 0
+    @correct_cards = []
+  end
+
+  def current_card
+    deck.cards[turns.count]
```
cjsim89, line 13: Even though Ruby lets us access instance variables without using an `@`, I recommending using it because it eliminates confusion as to where the program is getting that variable from. If you make an instance variable in the initialize method, it should have an `@` wherever it is called. Imagine this in a much bigger, more complex application - it could be really confusing really fast.

lib/deck.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  def initialize(cards)
+    @cards = cards
+    @count = cards.length
+  end
+
+  def cards_in_category(category)
+    #cards.select {|card| card.category == category}
+    card_category = []
+      cards.each do |card|
```
cjsim89, line 13: I left a similar comment elsewhere - if you reference an instance variable, it should have an `@` with it. This still works and Ruby is cool with it, it's more for readability and less confusion where possible.

flashcard_runner.rb
```diff
@@ -0,0 +1,45 @@
...
+    deck.cards.each do |card|
+      categories << card.category
+    end
+    categories.uniq.each do |category|
+      puts "#{category} - #{(round.percent_correct_by_category(category)).to_i}% correct"
+    end
+  end
+end
+
+new_game = FlashcardRunner.new
```
cjsim89, line 44: Great job putting your runner logic into a class. Next time, we will ask you to keep a class to its own file, and have a runner file be as small as possible, but this is a great step in the right direction.

lib/deck.rb
```diff
@@ -0,0 +1,21 @@
+require 'pry'
+class Deck
+  attr_reader :cards, :count, :cards_in_category
```
cjsim89, line 3: Don't need a `:cards_in_category` in the attr_reader, since it is already a method.

PR #406, opened 2022-05-15
flashcard project

KevinT001:
Struggling :\ Didn't complete

### DIFF
--- a/lib/card.rb
+++ b/lib/card.rb
@@ -1,0 +1,10 @@
+class Card
+
+  attr_reader :question, :answer, :category
+  def initialize (question, answer, category)
+    @question = question
+    @answer = answer
+    @category = category
+  end
+
+end
--- a/lib/deck.rb
+++ b/lib/deck.rb
@@ -1,0 +1,18 @@
+class Deck
+  attr_reader :cards
+  def initialize
+    @cards =["card_1", "card_2", "card_3", "card_4", "card_5", "card_6", "card_7", "card_8", "card_9"]
+  end
+
+  def draw
+    @cards.pop
+  end
+  
+  def shuffle
+    @cards.shuffle
+  end
+
+  def sort
+    @cards.sort
+  end
+end
--- a/lib/deck.rb
+++ b/lib/deck.rb
@@ -1,18 +1,7 @@
 class Deck
   attr_reader :cards
-  def initialize
-    @cards =["card_1", "card_2", "card_3", "card_4", "card_5", "card_6", "card_7", "card_8", "card_9"]
+  def initialize(cards = [])
+    @cards = cards
   end
 
-  def draw
-    @cards.pop
-  end
-  
-  def shuffle
-    @cards.shuffle
-  end
-
-  def sort
-    @cards.sort
-  end
 end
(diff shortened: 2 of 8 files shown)


lib/card.rb
```diff
@@ -0,0 +1,10 @@
+class Card
+
+  attr_reader :question, :answer, :category
+  def initialize (question, answer, category)
+    @question = question
+    @answer = answer
+    @category = category
+  end
+
+end
```

lib/deck.rb
```diff
@@ -0,0 +1,16 @@
...
+  def count
+    @cards.count
+  end
+
+  def cards_in_category(category)
+    cards.select do |card|
+      card.category == category
+    end
+  end
+end
```
megstang, line 16: Nice use of select here!

lib/round.rb
```diff
@@ -0,0 +1,9 @@
+class Round
+
+  attr_reader :deck, :turns
+  def initialize(deck)
+    @deck = deck
+    @turns = []
+  end
+
+end
```
megstang, line 9: What you have here looks good so far

lib/runner.rb
```diff
@@ -0,0 +1,14 @@
...
+
+card_2 = Card.new("Which continent is the coldest?", "Antartica", :Geography)
+
+card_3 = Card.new("Gotta catch'em all, is a phrase from what hit show originating in the mid 90's?", "Pokemon", :Popculture)
+
+
+card_4 = Card.new("The cartoon character 'Spongebob', lives in what kind of fruit?", "Pineapple", :Popculture)
+
+
+card_5 = Card.new("While the United States uses farenheit to label temperature measurements, what does majority of the world use?", "Celcius", :STEM)
```
megstang, line 14: Careful getting further ahead than what the iterations and interaction patterns guide you through. We want you to focus on the logic before focusing on how the game will run.

lib/turn.rb
```diff
@@ -0,0 +1,21 @@
...
+    @card = card
+  end
+
+  def correct?
+    if @guess == @card.answer
+      true
+    else
+      false
+    end
+  end
```
megstang, line 15: you could slim up this method by just having `@guess == @card.answer` since this will automatically return true or false for you.

spec/deck_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  it 'card count in deck' do
+    card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    card_2 = Card.new("The Viking spacecraft sent back to Earth photographs and reports about the surface of which planet?", "Mars", :STEM)
+    card_3 = Card.new("Describe in words the exact direction that  is 697.5° clockwise from due north?", "North north west", :STEM)
+
+    cards = [card_1, card_2, card_3]
+    deck = Deck.new(cards)
+    expect(cards.count).to eq(3)
```
megstang, line 32: with this assertion, you're not testing the cards attribute that's on your deck object. You're just testing the cards variable you have declared on line 30 which has nothing to do with your deck object. So, instead, you could write an assertion like this: 

`expect(deck.cards).to eq(cards)`

megstang (commented):
I left some feedback for you to look over. LEt me know if you have any questions. I'll follow up via slack with scores.

PR #407, opened 2022-05-15
Flashcards - Sunny Moore

sunny-moore:


lib/card.rb
```diff
@@ -0,0 +1,11 @@
...
+
+attr_reader :question, :answer, :category
+
+  def initialize(question, answer, category)
+    @question = question
+    @answer = answer
+    @category = category
+  end
+
+end
```

lib/deck.rb
```diff
@@ -0,0 +1,35 @@
+require 'pry'
+require './lib/card'
+class Deck
+  attr_accessor :cards, :cards_in_category
```
megstang, line 4: since you already have a cards_in_category method, you do not need to have an attr_Reader. We use attr_readers when we want to change attributes (instance variables in our initialize method) into methods so that we can read those attributes on that object outside of the class itself.

lib/deck.rb
```diff
@@ -0,0 +1,35 @@
...
+    #unsuccessful! Ask why
+    #try passing it a specific index
+    # @cards.each_with_index do |card, index|
+    #   if @cards[index].category == @category
+    #     @cards_in_category << @card[index]
+    #   end
+    # end
+    # return @cards_in_category
+
+  end
```
megstang, line 34: This works! But we can definitely slim it up a bit, and uncomplicate (is that a word?) some of your logic: 

```ruby
    cards_in_category = []
    @cards.each do |card|
      if card.category == category
           cards_in_category << card
      end
    end
    cards_in_category
```


There's a few things I did here .. instead of querying the cards array with an incremented index counter, I'm iterating through the cards collection since I know I want to look at every single one of them. Each-ing through the cards collection allows me to get to each card element in the array without having to manually look up the next index in the collection. 
I also am storing the cards_in_category array in a local variable instead of an instance variable. I'm doing this because I only need access to this variable within THIS method. We use instance variables (variables with the @ sign) when we want to access those values outside of the method that it was declared in.

lib/round.rb
```diff
@@ -0,0 +1,56 @@
+# require './lib/turn'
+
+class Round
+  attr_reader :turns, :number_correct
+  attr_accessor :deck
+  def initialize(deck)
+    @deck = deck
+    @turns = []
+    @number_correct = 0
+  end
```
megstang, line 10: since you don't need to access `number_correct` outside of this class, we don't need to set it as an attr_reader. Keep it as an instance variable in your initialize method though so you can reference it within this class in any method.

lib/round.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def current_card()
+    return @deck.cards[0]
+  end
+
+  def take_turn(guess)
+    # require 'pry'; binding.pry
+    this_turn = Turn.new(guess, current_card())
```
megstang, line 18: If you don't need to pass a method any arguments, you can leave off the () -- If you've had any experience in other programming languages prior to this, you might have noticed that it's common practice to put the () after method declarations, but in ruby, we don't need to do that if there aren't any arguments that we need to pass.

lib/round.rb
```diff
@@ -0,0 +1,56 @@
...
+    i = 0
+    while i < turns.length
+      if turns[i].card.category == category && turns[i].guess == turns[i].card.answer
+
+           number_correct_by_category +=1
+      end
+      i += 1
+    end
+    return number_correct_by_category
+  end
```
megstang, line 39: similarly here, you can iterate through the cards collection, rather than create your own incrementer to reference a specific part of the cards array.

lib/round.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def percent_correct_by_category(category)
+    correct = number_correct_by_category(category)
+    turns_in_category = 0
+    turns.each_with_index do |turn, index|
+      if turns[index].card.category == category
+        turns_in_category += 1
+      end
+    end
```
megstang, line 52: on line 49, you've written `turns[index]`. You actually have access to that turn you're trying to get, but it's in the variable `turn` (your first block variable).

lib/turn.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def initialize(guess, card)
+    @guess = guess
+    @card = card
+  end
+
+#refactor to not need a param
+  def correct?(guess)
+    @guess == card.answer
+
```
megstang, line 12: remove this blank line.

lib/turn.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+  def feedback()
```
megstang, line 14: keep a empty line between your methods.

megstang, line 14: you can take off the () here since this method doesn't take any arguments.

lib/turn.rb
```diff
@@ -0,0 +1,22 @@
+class Turn ()
+  attr_reader :guess, :card
+
+  def initialize(guess, card)
+    @guess = guess
+    @card = card
+  end
+
+#refactor to not need a param
+  def correct?(guess)
```
megstang, line 10: you do not need to pass this method an argument. You can see, even in the method, you don't reference `guess` but you reference `@guess`.. so, you can take off the placeholder here.

megstang (commented):
I left some technical feedback for you here. If you have any questions please let me know! I'll follow up tomorrow with your scores.

sunny-moore:
Thanks Meg. I forgot about the |index| portion of each and was struggling with making it work so I bet old school. Wish I’d left myself time for refactoring!

sunny-moore:
Hi Meg!

Any progress on score?  I'm on pins and needles!

Thanks,
Sunny

PR #408, opened 2022-05-15
Lucas Colwell

lcole37:


lib/card.rb
```diff
@@ -0,0 +1,12 @@
...
+              :answer,
+              :category
+
+  def initialize(question, answer, category)
+    @question = question
+    @answer = answer
+    @category = category
+  end
+
+end
```

lib/deck.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  def cards_in_category(topic)
+    array = []
+    @cards.each do |card|
+      if card.category == topic
+        array << card
+      end
+    end
+    return array
+  end
```
megstang, line 18: Nice work here, I'd only suggest naming the variable on line 11 something a bit more descriptive than `array`.

lib/round.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def current_card
+    @deck.cards[@turns.count]
+  end
+
+  def take_turn(answer)
+    new_turn = Turn.new(answer, @deck.cards[@turns.count])
+    @turns << new_turn
+    return new_turn
+  end
```
megstang, line 18: On line 11, when you have `@deck.cards[@turns.count]`, this is code that you have in your `current_card` method. So, you could replace that part of line 11 with a method call to `current_card` to dry up your code.

lib/round.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def number_correct
+    correct_count = 0
+    @turns.each do |turn|
+      if turn.feedback == "Correct!"
```
megstang, line 23: This will definitely work, you could also look at your `correct?` method since that will automatically return true or false for you

lib/turn.rb
```diff
@@ -0,0 +1,22 @@
...
+    @guess == @card.answer
+  end
+
+  def feedback
+    if @guess != @card.answer
+      "Incorrect."
+    else
+      "Correct!"
+    end
+  end
```
megstang, line 20: On line 15, this code is identical to the code you have in the `correct?` method. Because of that, you can replace line 15 to just read `if correct?`.

megstang (commented):
Nice work! I left some comments for you to look over. If you have any questions please let me know. I'll follow up later today with your scores via slack.

PR #409, opened 2022-05-15
Alex Mora

AlexMR-93:


lib/deck.rb
```diff
@@ -0,0 +1,17 @@
+class Deck
+attr_reader :cards
+  def initialize(cards)
+    @cards = cards
+
```
megstang, line 5: remove this extra line

lib/deck.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+    def count
+     @cards.length
+    end
```
megstang, line 9: this code is about 1 indent too far, have it line up with the last method, and have one line in between each method.

lib/deck.rb
```diff
@@ -0,0 +1,17 @@
...
+
+def cards_in_category(category)
```
megstang, line 11: indent this line

lib/deck.rb
```diff
@@ -0,0 +1,17 @@
...
+     cards.select do |card|
+       card.category == category
+     end
```
megstang, line 14: nice use of select

lib/deck.rb
```diff
@@ -0,0 +1,17 @@
...
+  def initialize(cards)
+    @cards = cards
+
+  end
+    def count
+     @cards.length
+    end
+
+def cards_in_category(category)
+     cards.select do |card|
```
megstang, line 12: instead of referencing the attr_reader for cards, call directly to the instance variable `@cards`. It takes out the middle man, and goes directly to the source of the value you're looking for.

lib/turn.rb
```diff
@@ -0,0 +1,26 @@
...
+  attr_reader :guess,:card
+
+  def initialize(guess,card)
+    @guess = guess
+    @card = card
+  end
+
+  def correct?
+    if @guess == card.answer
+    true
```
megstang, line 13: indent this line one more.

lib/turn.rb
```diff
@@ -0,0 +1,26 @@
...
+      else
+        false
+      end
```
megstang, line 16: indent these back one less

megstang, line 16: this method could be shortened to just one line: 

```
@guess == card.answer
```

becuase this will automatically return true or false. 

Also, I mentioned this before, but just to reiterate, I'd call to `@card` instead of `card`. Both will work, but you might as well call directly to the variable instead of to the method that returns that variable.

lib/turn.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def feedback
+    if correct? == true
```
megstang, line 20: This line could say `if correct?` since `correct?` will either be `true` or `false`.

lib/turn.rb
```diff
@@ -0,0 +1,26 @@
...
+      "Correct!"
+    else
+    "Incorrect."
```
megstang, line 23: indent this in

spec/round_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    expect(round).to((be_instance_of(Round)))
+  end
+
+  it("has turns") do
+    card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    card_2 = Card.new("The Viking spacecraft sent back to Earth photographs and reports about the surface of which planet?", "Mars", :STEM)
+    card_3 = Card.new("Describe in words the exact direction that is 697.5° clockwise from due north?", "North north west", :STEM)
+    deck = [card_1, card_2, card_3]
+    round = Round.new(deck)
+    expect(round.turns).to(eq(turns))
```
megstang, line 23: you don't have a variable or method called `turns` so i imagine this line would error out. If no turns have been taken, we could write the assertion like this: 

`expect(round.turns).to eq([])`

spec/turn_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+       card = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+       turn = Turn.new("Juneau", card)
+       expect(turn.feedback).to eq("Correct!")
+     end
+
+     it "has card w. answer equal to mercury" do
+        card = Card.new("Which planet is closest to the sun?", "Mercury", :STEM)
+        turn = Turn.new("Saturn",card)
+        expect(turn.card).to eq(card)
+      end
```
megstang, line 42: also test for if a incorrect guess is taken -- make sure `correct?` returns false, and that the feedback is "Incorrect".

megstang (commented):
I left some feedback for you on this PR. You can expect scores to be sent out tomorrow. Please let me know if you have any questions.

PR #410, opened 2022-05-15
Robert Schlereth

rschlereth:


lib/card.rb
```diff
@@ -0,0 +1,9 @@
+class Card
+  attr_reader :question, :answer, :category
+
+ def initialize(question, answer, category)
+   @question = question
+   @answer = answer
+   @category = category
+ end
+end
```

lib/deck.rb
```diff
@@ -0,0 +1,30 @@
...
+  #   require "pry"; binding.pry
+    @cards.count
+  # end
+  # require "pry"; binding.pry
+  end
+  def cards_in_category(category)
+    # require "pry"; binding.pry
+    @cards.select do |card|
+      card.category == category
+  end
```
megstang, line 19: this a great use of select, nice work!

lib/deck.rb
```diff
@@ -0,0 +1,30 @@
...
+    # matching_cards = []
+    # @cards.each do |card|
```
megstang, line 21: Just as a general note, clean up your comments/binding.prys/code attempts before pushing up to main. It's okay to leave helpful comments for yourself, but it looks like you ended up commenting out old code which you didn't use, and you can go ahead and clean that up before merging into main.

lib/round.rb
```diff
@@ -0,0 +1,38 @@
+class Round
+  attr_reader :deck, :turns
+  def initialize(deck)
+    @deck = deck
+    @turns = []
+    @number_correct = 0
+    @number_wrong = 0
+  end
```
megstang, line 8: I like seeing that you recognize that you don't need an attr reader for every attribute you define -- that's a misconception I see a lot of students have at the beginning of mod 1, and looks like you are understanding that you only need an attr_Reader for an attribute when you want to read it outside of this class.

lib/round.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def current_card
+    @deck.cards.first
+  end
+
+  def take_turn(user_input)
+    turn = Turn.new(user_input, current_card)
```
megstang, line 15: nice job calling to your already written method, `current_card`.

lib/round.rb
```diff
@@ -0,0 +1,38 @@
+class Round
+  attr_reader :deck, :turns
+  def initialize(deck)
+    @deck = deck
+    @turns = []
+    @number_correct = 0
+    @number_wrong = 0
```
megstang, line 7: I'm curious why you decided to creat this number_wrong variable.. I don't see a need for it, but curious what your plan was for it.

lib/round.rb
```diff
@@ -0,0 +1,38 @@
...
+    else
+      @number_wrong += 1
+    end
+    @turns << turn
+    turn
+  end
+
+  def number_correct
+    @number_correct
+  end
```
megstang, line 27: If you had an attr_reader for number_correct, you would not need this method. Either is fine with me!

lib/turn.rb
```diff
@@ -0,0 +1,23 @@
...
+    @card = card
+
+  end
+
+  def correct?
+    @guess == @card.answer
+  end
+
+  def feedback
+    if correct? == true
```
megstang, line 15: since correct? will evaluate to true or false, you don't need the comparison operator here. You could have this line just read `if correct?`

spec/turn_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  it "can return correct or incorrect" do
+    card = Card.new("Which planet is closest to the sun?", "Mercury", :STEM)
+    turn = Turn.new("Saturn", card)
+    expect(turn.correct?).to be_falsey
+  end
+  it "can return correct or incorrect" do
+    card = Card.new("Which planet is closest to the sun?", "Mercury", :STEM)
+    turn = Turn.new("Mercury", card)
+    expect(turn.correct?).to be_truthy
+  end
```
megstang, line 39: careful with this `be_truthy` and `be_falsey` assertions. In ruby, all things are 'truthy' as long as they don't evaluate to `false` or `nil`. So, it's possible that this assertion can lead to falsley passing tests.

megstang (commented):
I left some comments for you to review. Please let me know if you have any questions! I'll follow up in slack later today with your scores.

PR #411, opened 2022-05-15
Wesley Miranda

wes30303:


spec/trun_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  it 'has correct' do
+    card = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    turn = Turn.new("Juneau", card)
+    expect(turn.correct?).to eq(true)
+  end
+
+  it 'has feedback' do
+    card = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    turn = Turn.new("Juneau", card)
+    expect(turn.feedback).to eq("Correct")
```
epintozzi, line 32: This method in the `Turn` class returns "Correct." with a period at the end, so as a result this test fails because it doesn't match exactly. Tests looks for very specific outputs, so we just need to be extra careful to make sure it matches what the test says.

spec/round.rb
```diff
@@ -0,0 +1,55 @@
+class Round
```
epintozzi, line 1: I think this probably got put in the `spec` folder by accident but the `Round` class should be up in `lib`

lib/card.rb
```diff
@@ -0,0 +1,8 @@
+class Card
+attr_reader :question, :answer, :category
+    def initialize(question, answer, category)
+      @question = question
+      @answer = answer
+      @category = category
+    end
```
epintozzi, line 7: This is a little bit nit picky but is the kind of thing that will stand out when you get to code challenges and things like that on the job hunt. Be careful to watch your indentation to make sure it is consistent and matches ruby convention. Here, we would want line 2 indented two spaces. All our method `def`s should also be indented two spaces, as well. We also want an empty line after our `attr_reader`. Like this:

```ruby
class Card
  attr_reader :question, :answer, :category

  def initialize(question, answer, category)
    @question = question
    @answer = answer
    @category = category
  end
end
```

spec/card_spec.rb
```diff
@@ -24,4 +26,18 @@
...
+    expect(card.question).to eq("What is the capital of Alaska?")
+  end
+
+
+
+
+
+
+
+
```
epintozzi, line 42: It's good practice to get rid of extra lines and spaces like this before committing our code.

spec/card_spec.rb
```diff
@@ -24,4 +26,18 @@
 
     expect(card.category).to eq(:Geography)
   end
+
+  it 'exists' do
+    card = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+
+    expect(card.question).to eq("What is the capital of Alaska?")
+  end
```
epintozzi, line 34: This test is the same as the test on line 11 so we can get rid of it.

lib/deck.rb
```diff
@@ -0,0 +1,21 @@
...
+    end
+
+    def cards_in_category (category)
+      cards_in_category = []
+        @cards.each do |card|
+          if card.category == category
+            cards_in_category << card
+          end
+        end
+        return cards_in_category
```
epintozzi, line 19: This works great! As you get more comfortable with enumerables, you can start experimenting with different options. Here's another one that works well in this scenario:
```ruby
    cards.select do |card|
      card.category == category
    end
```

spec/deck_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+    card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    card_2 = Card.new("The Viking spacecraft sent back to Earth photographs and reports about the surface of which planet?", "Mars", :STEM)
+    card_3 = Card.new("Describe in words the exact direction that is 697.5° clockwise from due north?", "North north west", :STEM)
+    cards = [card_1, card_2, card_3]
+    deck = Deck.new(cards)
+
+    expect(deck.cards).to eq([card_1, card_2, card_3])
+    #require 'pry' ; binding.pry
+
+
```
epintozzi, line 19: Extra line breaks are not needed here

lib/turn.rb
```diff
@@ -0,0 +1,24 @@
+
```
epintozzi, line 1: Extra lines on line 1, 5, 22, 23 can be removed. Also double check the indentation for the `correct?` and `feedback` methods.

lib/turn.rb
```diff
@@ -0,0 +1,24 @@
...
+class Turn
+attr_reader :guess, :card, :feedback
```
epintozzi, line 3: We don't need a `feedback` `attr_reader` because you also define a `feedback` method on line 15. `attr_readers` are just shortcuts that allow us to skip manually defining a method if we only need to return the value of an instance variable.

spec/round.rb
```diff
@@ -0,0 +1,55 @@
...
+
+attr_reader :turns, :deck, :turn, :number_correct, :take_turn, :cards, :number_correct_by_category
```
epintozzi, line 3: `attr_readers` are used to allow us read access to the value stored on instance variables. This means the only `attr_readers` we need for this class are `:turns, :deck, :number_correct`

spec/round_spec.rb
```diff
@@ -0,0 +1,229 @@
+require './lib/turn'
+require './lib/card'
+require './lib/deck'
+require './lib/round'
+require 'rspec'
+
+
+RSpec.describe Deck do
```
epintozzi, line 8: This should be `Round` not `Deck`

spec/round_spec.rb
```diff
@@ -0,0 +1,229 @@
...
+    deck = Deck.new([card_1, card_2, card_3])
+    round = Round.new(deck)
+    new_turn = round.take_turn("Juneau")
+
+
+    expect(new_turn.correct?).to eq(true)
+  end
+
+
+  it 'has an answer' do
```
epintozzi, line 69: A better name for this test might be something like `it 'adds new turn to turns'`

spec/round.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  def take_turn(guess)
+    turn = Turn.new(guess, current_card)
+      @turns << turn
+      if turn.correct?
+        @number_correct += 1
+      end
+      deck.cards.rotate!
+    return turn
+  end
```
epintozzi, line 24: Several of the methods in this class are a little off for indentation. This method should look more like this, for example:
```ruby
  def take_turn(guess)
    turn = Turn.new(guess, current_card)
    @turns << turn
    if turn.correct?
      @number_correct += 1
    end
    deck.cards.rotate!
    turn
  end
```

flashcard_runner.rb
```diff
@@ -0,0 +1,31 @@
...
+card_1 = Card.new("What is 5 + 5?", "10", :STEM)
+card_2 = Card.new("What is Rachel's favorite animal?", "Rat", :TuringStaff)
+card_3 = Card.new("What is Mike's middle name?", "nobody knows", :TuringStaff)
+card_4 = Card.new("What cardboard cutout lives at Turing?", "Justin Bieber", :PopCulture)
+deck = Deck.new([card_1, card_2, card_3, card_4])
+round = Round.new(deck)   
+
+
+def start(round)
+  p "Welcome! You're playing with 4 cards."
```
epintozzi, line 17: To make this a little more dynamic, you could interpolate the number of cards with something like this, in case there are ever more than 4 cards:
`"Welcome! You're playing with #{round.deck.count} cards."`

epintozzi (commented):
Good work Wes! Overall this works well. The functionality is correct. Pay close attention to your indentation and extra lines. Cleaning that up will make your code easier to read and a little more conventional. If you'd like help finding some resources on this, let me know! I'll follow up with your scores in slack by end of day Friday.

PR #412, opened 2022-05-15
Thiago Silveira

tig-o:


cjsim89 (commented):
@tig-o Great work on this project! I like your use of enumerables to keep your code concise, and the runner file is also nicely compacted and uses the classes well. If you have time later, try to finish out iteration 4 to have this be an even better portfolio piece.

PR #413, opened 2022-05-15
Project 

SandyyMarie:
Only couldn't complete the right syntax for txt file input

### DIFF
--- a/lib/card.rb
+++ b/lib/card.rb
@@ -1,0 +1,11 @@
+class Card
+  attr_reader :question,:answer, :category, :prompt
+
+
+  def initialize(question, answer, category) #initializes the card class with basic properties
+    @question = question
+    @answer = answer
+    @category = category #looks like an attr_reader but is like a string variable sorta, would have to utilize ._s to manipulate
+  end
+
+end
--- a/lib/card.rb
+++ b/lib/card.rb
@@ -1,6 +1,7 @@
 class Card
   attr_reader :question,:answer, :category, :prompt
 
+  cards = [] #not sure if to init cards here or in deck
 
   def initialize(question, answer, category) #initializes the card class with basic properties
     @question = question
--- a/lib/card.rb
+++ b/lib/card.rb
@@ -1,9 +1,10 @@
 class Card
   attr_reader :question,:answer, :category, :prompt
 
-  cards = [] #not sure if to init cards here or in deck
+
 
   def initialize(question, answer, category) #initializes the card class with basic properties
+    cards = [] #not sure if to init cards here or above
     @question = question
     @answer = answer
     @category = category #looks like an attr_reader but is like a string variable sorta, would have to utilize ._s to manipulate
--- a/lib/card_generator.rb
+++ b/lib/card_generator.rb
@@ -1,0 +1,17 @@
+require './card' #to call card class to set classes
+
+filename = "cards.txt"
+class CardGenerator
+
+  def initialize(filename)
+    # file = File.open(filename)
+    # array = File.read(filename).split(",")
+    # cards = Card.new(array)
+
+    IO.readlines(filename).each do |line|
+      values = []
+      values = line.split(",")   #.collect(&:strip)
+      cards = Card.new(values[0],values[1],values[2])
+    end
+  end
+end
--- a/lib/deck.rb
+++ b/lib/deck.rb
@@ -1,30 +1,24 @@
-# Create a Deck class with an accompanying test file. A Deck is initialized with an array of Card objects. A Deck should also be able to return cards based on a given category.
 class Deck
   attr_reader :cards, :count
 
   def initialize(cards)
     @cards = cards
-    @count = @cards.count() #not sure if should make individual method - .count() counts the # of objects in array
+    @count = @cards.count()
   end
 
   def cards_in_category(selected_category) #selected category = category you're searching extend for
      #setting iterate# to current count will (if i == @count, break)
 
     current_search = []
-    # couldn't figure out how to use  do
-    # @cards.each do |card|
-    #   if @cards.category == selected_category
-    #     current_search << card
-    #   end
 
     i = 0
-    @count.times do #will iterate through each card category and add to current search to print
+    @count.times do
       if @cards[i].category == selected_category
         current_search << @cards[i]
       i += 1
     end
 
-      return current_search
+      return current_search #will iterate through each card category @count times and if it finds a card in that category it adds it to current_search to return array after.
     end
 
 
--- a/lib/deck.rb
+++ b/lib/deck.rb
@@ -5,9 +5,28 @@ class Deck
   def initialize(cards)
     @cards = cards
     @count = @cards.count() #not sure if should make individual method - .count() counts the # of objects in array
-
   end
 
+  def cards_in_category(selected_category) #selected category = category you're searching extend for
+     #setting iterate# to current count will (if i == @count, break)
+
+    current_search = []
+    # couldn't figure out how to use  do
+    # @cards.each do |card|
+    #   if @cards.category == selected_category
+    #     current_search << card
+    #   end
+
+    i = 0
+    @count.times do #will iterate through each card category and add to current search to print
+      if @cards[i].category == selected_category
+        current_search << @cards[i]
+      i += 1
+    end
+
+      return current_search
+    end
 
 
+  end
 end
--- a/lib/deck.rb
+++ b/lib/deck.rb
@@ -2,7 +2,7 @@ class Deck
   attr_reader :cards, :count
 
   def initialize(cards)
-    @cards = cards
+    @cards = cards #initializing cards array
     @count = @cards.count()
   end
 
--- a/lib/deck.rb
+++ b/lib/deck.rb
@@ -1,0 +1,13 @@
+# Create a Deck class with an accompanying test file. A Deck is initialized with an array of Card objects. A Deck should also be able to return cards based on a given category.
+class Deck
+  attr_reader :cards, :count
+
+  def initialize(cards)
+    @cards = cards
+    @count = @cards.count() #not sure if should make individual method - .count() counts the # of objects in array
+
+  end
+
+
+
+end
--- a/lib/round.rb
+++ b/lib/round.rb
@@ -5,20 +5,21 @@
 # The take_turn method is the crux of this problem. The take_turn method takes a string representing the guess. It should create a new Turn object with the appropriate guess and Card. It should store this new Turn, as well as return it from the take_turn method. Also, when the take_turn method is called, the Round should move on to the next card in the deck.
 
 class Round
-  attr_reader :deck, :turns
+  attr_reader :deck, :turns, :current_card
 
   def initialize(deck)
     @deck = deck #deck of cards for this round
     @turns = [] #initialized with empty array
+    @current_card = @deck.cards[0] #initializes current card with first card of deck
 
   end
 
-  def take_turn(guess_string)
-
-    #should create a new turn object with guess + card, and store in new turn
-
-    #return new Turn
-
-  end
+  # def take_turn(guess_string)
+  #   #shovel guesses into turns array
+  #   #should create a new turn object with guess + card, and store in new turn
+  #
+  #   #return new Turn
+  #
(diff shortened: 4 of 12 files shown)


SandyyMarie:
Closing request

SandyyMarie:
Not sure if I should be closing these myself, reopening!

lib/card.rb
```diff
@@ -0,0 +1,13 @@
+class Card
+  attr_reader :question,:answer, :category, :prompt
```
epintozzi, line 2: I'm not sure what `:prompt` is for. Maybe leftover from something you were trying out? In any case, it doesn't appear to be used for anything so it can/should be removed.
```suggestion
  attr_reader :question, :answer, :category
```

lib/card.rb
```diff
@@ -0,0 +1,13 @@
...
+
+
+
```
epintozzi, line 5:
```suggestion
```

It's good practice to remove extra line breaks in our code. It's cleaner and easier to read.

lib/card.rb
```diff
@@ -0,0 +1,13 @@
...
+  def initialize(question, answer, category) #initializes the card class with basic properties
+    cards = [] #not sure if to init cards here or above
```
epintozzi, line 7: This isn't being used for anything so can be removed. It looks like you were wondering if this should be here or not. We would want to hold a collection of cards in a `Deck` but not within a card itself.
```suggestion
```

lib/card.rb
```diff
@@ -0,0 +1,13 @@
...
+    @question = question
+    @answer = answer
+    @category = category #looks like an attr_reader but is like a string variable sorta, would have to utilize ._s to manipulate
```
epintozzi, line 10: In the project spec, category is being sent is as a symbol. Depending on what you need/want to do with it, it does not necessarily have to be converted to a string first.

lib/deck.rb
```diff
@@ -0,0 +1,26 @@
+class Deck
+  attr_reader :cards, :count
+
+  def initialize(cards)
+    @cards = cards #initializing cards array
+    @count = @cards.count()
```
epintozzi, line 6: This isn't something we really need to instantiate with. It's more of a helper method and can live in that class as such. Something like this would be sufficient and then would also not need an `attr_reader` either.
```ruby
def count
  @cards.count
end
```

spec/deck_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  it 'will return cards of given category' do #done
+    card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    card_2 = Card.new("The Viking spacecraft sent back to Earth photographs and reports about the surface of which planet?", "Mars", :STEM)
+    card_3 = Card.new("Describe in words the exact direction that is 697.5° clockwise from due north?", "North north west", :STEM)
+    cards = [card_1, card_2, card_3]
+
+    deck = Deck.new(cards)
+
+    expect(deck.cards_in_category(:STEM)).to eq([]) #returns array of stem cards?
```
epintozzi, line 49: This expectation should be looking for a collection of cards, not an empty array if it's working correctly. 
```suggestion
    expect(deck.cards_in_category(:STEM)).to eq([card_1, card_2]) #returns array of stem cards?
```

lib/deck.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    i = 0
+    @count.times do
+      if @cards[i].category == selected_category
+        current_search << @cards[i]
+      i += 1
+    end
+
+      return current_search #will iterate through each card category @count times and if it finds a card in that category it adds it to current_search to return array after.
+    end
```
epintozzi, line 22: There's a little bit to unpack here. Please try to watch your extra spaces and indentations. It can make code difficult to read when it's not aligned correctly. It's a minor thing but is the kind of stuff that will be noticed on the job hunt.

Second, this method _almost_ works. Right now, your counter is inside the `if @cards[i].category == selected_category` which means if the categories *DON'T* match, then the counter won't increase, so we should move the counter outside of the if statement.

Next, `return current_search` is happening within the `times` loop. A `return` stops execution of code which means this loop will stop after the first time through because when it hits that `return`, the method ends. The `return current_search` should be moved outside of the `times` loop. And as a fun fact, ruby implicitly returns that resulting value of the last line of code in a method so we don't have to explicitly tell it to `return` unless we want the code to stop executing early (which happens but isn't super often). 

In case I lost you there, rewritten, it would look something like this:
```ruby
  def cards_in_category(selected_category)
    current_search = []
    i = 0
    @count.times do
      if @cards[i].category == selected_category
        current_search << @cards[i]
      end
      i += 1
    end
    current_search
  end
```

However, I'm going to go one step further and suggest that ruby has some nice built in enumerables that make this who calculation a lot simpler. We could also write this as:
```ruby
    cards.select do |card|
      card.category == category
    end
  end
```

lib/turn.rb
```diff
@@ -0,0 +1,36 @@
+class Turn
+  attr_reader :guess, :current_card, :wrong_or_right
```
epintozzi, line 2: We generally only use `attrs` with instance vars, so `wrong_or_right` should be removed.
```suggestion
  attr_reader :guess, :current_card
```

lib/turn.rb
```diff
@@ -0,0 +1,36 @@
...
+      @current_card = current_card
+
+      # @wrong_or_right = @current_card.answer.include?(@guess)
+      # so I need to have an array or if block of available questions and what their answers are
+
+  end
+
+  def card
+    @current_card #implicit return of current_turn which is a string right now
+  end
```
epintozzi, line 21: I would probably just call this `@card = card` and then also use that same name for the `attr_reader`. That way you can avoid this scenario where you've essentially just written an alias for `current_card` called `card`.

spec/turn_spec.rb
```diff
@@ -0,0 +1,40 @@
+require './lib/card'
+require './lib/turn'
+require './lib/deck'
+require './lib/round'
+require 'rspec'
+require 'pry'
+
+RSpec.describe Turn do #done
+  it 'exists' do
+    current_turn = Turn.new("This is the answer", "This might be the question?")
```
epintozzi, line 10: `Turn` takes in a guess at the answer and the `Card` object that is being guessed upon, so this spec (and all the ones following) should be setup more like this:

```ruby
  it 'exists' do
    card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
    current_turn = Turn.new("Juneau", card_1)

    expect(current_turn).to be_instance_of(Turn)
  end
```

lib/turn.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def correct?
+    @wrong_or_right = @current_card.answer.include?(@guess)
```
epintozzi, line 24:
```suggestion
@current_card.answer.include?(@guess)
```

`@current_card.answer.include?(@guess)` returns true on its own so we don't also need to set it equal to an instance variable. We can call the `correct?` method directly and get the same answer.

lib/turn.rb
```diff
@@ -0,0 +1,36 @@
...
+    # = @answer.include?(@guess)
+  end
+
+  def feedback
+    if @wrong_or_right == TRUE
+      "Correct!" #implicit return?? (otherwise nill or doubled)
+    elsif @wrong_or_right == FALSE
+      "Incorrect." #implicit return
+    end
```
epintozzi, line 33: Because `correct?` answers the question of `@wrong_or_right == TRUE` we can instead rewrite this method like this:

```ruby
  def feedback
    if correct?
      "Correct!"
    else
      "Incorrect."
    end
  end
```

spec/turn_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  #
+  #   expect(current_turn.correct?).to eq(TRUE)
+  # end
+  #
+  # it 'returns "Correct!"' do
+  #   current_turn = Turn.new("This is the answer", "This might be the question?")
+  #
+  #   expect(current_turn.feedback).to eq("Correct!")
+  # end
+
```
epintozzi, line 38: It's also a good idea to test the false/negative values as well so test that `correct?` can return a `false` accurately and that `feedback` will return `Incorrect.` accurately.

spec/round_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+  it 'returns current turns/guesses' do #done
+    card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    card_2 = Card.new("The Viking spacecraft sent back to Earth photographs and reports about the surface of which planet?", "Mars", :STEM)
+    card_3 = Card.new("Describe in words the exact direction that is 697.5° clockwise from due north?", "North north west", :STEM)
+    deck = Deck.new([card_1, card_2, card_3])
+    round = Round.new(deck)
+
+    expect(round.turns).to eq([])
+  end
+
```
epintozzi, line 28: We should also test that `round.deck` returns the `Deck`

lib/round.rb
```diff
@@ -0,0 +1,78 @@
...
+    @number_correct = 0 #inititalized with 0
+    @correct_qs = [] #placeholder array for correct answers
+
+  end
+
+
+  def current_card #moved to individual method to make sure it updated the iteration
+    @current_card = @deck.cards[@current_turn]#initializes current card with first card of deck
+    return @current_card
+  end
```
epintozzi, line 17: This works! Something else that might be a little more straightforward and easier to read would be something like this:
```ruby
def current_card
    deck.cards.first
  end
```

And then physically rotate/shuffle the deck after the user takes a turn. That way, you don't also have to keep track of the `@current_turn`.

```ruby
  def rotate_deck
    deck.cards.rotate!
  end
```

lib/round.rb
```diff
@@ -0,0 +1,78 @@
...
+        if @turns[i].correct? #turns[i] returning (#stuff)
+          @num_right += 1
+        end
+      end
+      i += 1
+    end
+
+    return @num_right
+  end
+
```
epintozzi, line 58: The `Turn` also holds the cards used (and thus the category) so you make be able to consolidate what's happening here a little bit with something more like:
```ruby
def number_correct_by_category(category)
    correct = 0
    turns.each do |turn|
      correct += 1 if turn.card.category == category && turn.correct?
    end
    correct
  end
```

lib/round.rb
```diff
@@ -0,0 +1,78 @@
+class Round
+  attr_reader :deck, :turns, :current_card, :current_turn, :number_correct,:cat_choice_qs, :correct_qs, :num_right
```
epintozzi, line 2: There are a lot of `attr_reader`s here. We should be able to get away with just a few and use our methods for handling the rest of our data.

epintozzi (commented):
Overall great effort! I left some specific feedback on how this code can be streamlined a bit. Some high level thoughts - it's good practice to keep an eye on extra line breaks and indentation so that it's easy to read. Keep practicing your ruby enumerable methods. There are several cases where you're looping through things that ruby has some nice built in methods for. It takes practice to get familiar with those options, but some of these things will feel a lot easier to solve once you start getting comfortable with it. 

It's also pretty conventional in ruby to _not_ comment your code. The idea is that if you are writing clean and easy to read code, it shouldn't need comments to be understood. Other programming languages prefer the commenting convention but it's not something you'll see a ton of in ruby. For final project submissions, take a few minutes at the end to clean up extra comments and trailing lines/spaces, and everything will feel much more organized. Well send scores via Slack by end of day Friday.

lib/card_generator.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  def initialize(filename)
+    # file = File.open(filename)
+    # array = File.read(filename).split(",")
+    # cards = Card.new(array)
+
+    IO.readlines(filename).each do |line|
+      values = []
+      values = line.split(",")   #.collect(&:strip)
+      cards = Card.new(values[0],values[1],values[2])
```
epintozzi, line 14: This looks super close! I think if you change cards in line 14 to an array and do something like:
`cards << Card.new(values[0],values[1],values[2])` this should work!

Pop a pry in between lines 11&12 and verify how the lines are coming in, what they look like split, etc. When I do that, it looks how I'd expect it to and can generate a card!

PR #414, opened 2022-05-15
Wyatt Kribs

WyattKribs:


lib/card.rb
```diff
@@ -0,0 +1,8 @@
+class Card
+  attr_reader :question, :answer, :category
+  def initialize(question, answer, category)
+    @question = question
+    @answer = answer
+    @category = category
+  end
+end
```
megstang, line 8: 👍

lib/deck.rb
```diff
@@ -0,0 +1,24 @@
...
+  def count
+    cards.size
+  end
+
+  def cards_in_category(thing)
+    card_category = []
+    cards.each do |card|
+      if card.category == thing
+        card_category << card
+      else
```
megstang, line 16: no need for an `else` if there is not else condition that you need to take care of.

lib/round.rb
```diff
@@ -0,0 +1,31 @@
+require './lib/turn'
+class Round
+  attr_reader :deck, :turns, :new_turn, :number_correct, :number_of_turns
+  def initialize(deck)
+    @deck = deck
+    @turns = []
+    @number_correct = 0
+    @number_of_turns = 0
+  end
```
megstang, line 9: you only need attr_readers for the attributes that you want to be able to read outside of this class. So, for that reason, you only need to have attr readers for deck and turns.

lib/round.rb
```diff
@@ -0,0 +1,31 @@
...
+  #note to self, practice instanciating classes in methods
+  def take_turn(guess)
+    turns.unshift(deck.cards[0])
+    deck.cards.shift
+    @new_turn = Turn.new(guess, current_card)
+    if current_card.answer == guess
+      add_correct
+    end
+    new_turn
+  end
```
megstang, line 21: you do not need to access `@new_turn` outside of this method, so this variable should be a local variable instead of an instance variable.

lib/round.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def current_card
+    turns[0]
+  end
```
megstang, line 25: this will return the first turn, but not the current card.

lib/round.rb
megstang: you'll also need to add the new turn into the turn collection

lib/turn.rb
```diff
@@ -0,0 +1,24 @@
...
+    @card = card
+  end
+
+  def correct?
+    if guess == card.answer
+      true
+    else
+      false
+    end
+  end
```
megstang, line 14: you can simplify this method to just one line: 


```
  def correct?
     guess == card.answer
  end
```

since == will either return true or false automatically.

lib/turn.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def feedback
+    if correct? == true
```
megstang, line 17: since `correct?` will return true or false, you do not need to have the `==true` on this line.

spec/deck_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    expect(@deck.cards).to eq([@card_1, @card_2, @card_3])
+  end
+
+  it "can tell you how many cards there are" do
+    expect(@deck.count).to eq(3)
+  end
+
+  it "can tell you which cards are in a catagory" do
+    expect(@deck.cards_in_category(:STEM)).to eq([@card_2, @card_3])
+  end
```
megstang, line 26: test more than just one case for this method -- make sure it works for other categories passed in.

megstang (commented):
Let me know if you have any questions on my feedback. I'll follow up tomorrow with scores.

PR #417, opened 2022-05-19
Candace Eckels

candace-eckels:
I misunderstood the directions

### DIFF
--- a/lib/card.rb
+++ b/lib/card.rb
@@ -1,0 +1,11 @@
+require 'pry'
+
+class Card
+    attr_reader :question, :answer, :category
+    def initialize(question,answer,category)
+        @question = question
+        @answer = answer
+        @category = category
+    end
+
+end
--- a/lib/deck.rb
+++ b/lib/deck.rb
@@ -1,0 +1,10 @@
+require 'pry'
+
+class Deck 
+    attr_reader :cards
+    def initialize(cards)
+        @cards = []
+    end
+
+
+end
--- a/lib/deck.rb
+++ b/lib/deck.rb
@@ -1,10 +1,22 @@
 require 'pry'
 
 class Deck 
-    attr_reader :cards
+    attr_reader :cards, :count
     def initialize(cards)
-        @cards = []
+        @cards = cards
     end
 
+    def count
+        @cards.length
+    end
 
+    def cards_in_category(category)
+        card_same_category = []
+            @cards.each do |card|
+                if card.category == category
+                    card_same_category << card
+                end
+            end
+        card_same_category
+    end
 end
--- a/lib/deck.rb
+++ b/lib/deck.rb
@@ -2,13 +2,9 @@ require 'pry'
 
 class Deck 
     attr_reader :cards, :count
-    def initialize(cards)
-        @cards = cards
-    end
+    def initialize(cards); @cards = cards end
 
-    def count
-        @cards.length
-    end
+    def count; @cards.length end
 
     def cards_in_category(category)
         card_same_category = []
--- a/lib/deck.rb
+++ b/lib/deck.rb
@@ -1,7 +1,8 @@
 require 'pry'
 
 class Deck 
-    attr_reader :cards, :count
+    attr_reader :count
+    attr_accessor :cards
     def initialize(cards); @cards = cards end
 
     def count; @cards.length end
--- a/lib/deck.rb
+++ b/lib/deck.rb
@@ -1,8 +1,8 @@
 require 'pry'
 
 class Deck 
-    attr_reader :count
     attr_accessor :cards
+
     def initialize(cards); @cards = cards end
 
     def count; @cards.length end
--- a/lib/round.rb
+++ b/lib/round.rb
@@ -4,6 +4,7 @@ class Round
     attr_reader :deck, :turns
     def initialize(deck)
         @deck = deck
+        @turns = []
     end
 
 end
--- a/lib/round.rb
+++ b/lib/round.rb
@@ -35,8 +35,9 @@ class Round
         @correct_by_category
     end 
 
-    def percent_correct
-        (number_correct / @turns.count.to_f)*100
-    end
+    def percent_correct; (number_correct / @turns.count.to_f)*100 end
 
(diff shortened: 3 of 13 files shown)


spec/turn_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    it 'can return feedback if answer is correct/incorrect' do
+        @turn1 = Turn.new("Juneau", @card)
+
+        expect(@turn.feedback).to eq('Incorrect.')
+        expect(@turn1.feedback).to eq('Correct!')
+    end
+
+    it 'can evaluate if the answers match but casing is off' do
```
abdulredd, line 31: Really impressed by this test that checks for casing.

spec/turn_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+        expect(@turn).to be_a(Turn)
+    end
+
+    it 'has attributes' do
+        expect(@turn.guess).to eq('Saturn')
+        expect(@turn.card).to eq(@card)
+    end
+
+    it 'can evaluate if the guess is correct' do
+        expect(@turn.correct?).to eq false
```
abdulredd, line 21: Would have liked to see another test that checked for a `true` output as well

lib/card.rb
```diff
@@ -0,0 +1,11 @@
+require 'pry'
```
abdulredd, line 1: Don't forget to remove your `require 'pry'` lines from your code.

lib/deck.rb
```diff
@@ -0,0 +1,19 @@
+require 'pry'
+
+class Deck 
+    attr_accessor :cards
+
+    def initialize(cards); @cards = cards end
+
+    def count; @cards.length end
```
abdulredd, line 8: While this single line method is cute, however, I would recommend using the more verbose syntax instead.

Stylistic choices are often personal preference so you can take this advice with a grain of salt.

lib/round.rb
```diff
@@ -0,0 +1,78 @@
...
+        @turns.each do |turn|
+            if turn.card.category == input
+                total_by_category += 1 
+            end
+        end
+        (number_correct_by_category(input) / total_by_category.to_f)*100
+    end
+
+    def start
+        puts "Welcome! You're playing with 4 cards. \n--------------------------------------"
```
abdulredd, line 54: Even though when running your application with `ruby lib/runner.rb` there are 8 cards added to the deck, because this line is hard-coded, it always prints `"Welcome! You're playing with 4 cards."` to the console.

A better choice would have been to use some interpolation here that outputted the actual number of cards in the deck.

`"Welcome! You're playing with #{@deck.cards.count} cards.`

lib/round.rb
```diff
@@ -0,0 +1,78 @@
...
+            guessing = gets.chomp
+            take_turn(guessing)
+            puts turns.last.feedback
+        end
+        if turns.count == deck.count
+            puts "****** Game over! ******"
+            puts "You had #{number_correct} correct guesses out of #{card_count} for a total score of #{percent_correct.to_i}%."
+            turns_correct = []
+            turns.each do |turn|
+                if turns_correct.include?(turn.card.category) == false
```
abdulredd, line 69: You don't necessarily have to do the comparison to false, you could also rewrite this line to something like this:

```ruby
if !turns_correct.include?(turn.card.category) #<=== notice the "not" operator or bang in the front of turns_correct. This can be read like if turns_correct dot not include turn.card.category then do this...
  turns_correct << turn.card.category
end

#[REST OF CODE...]
```

lib/round.rb
```diff
@@ -0,0 +1,78 @@
...
+                    @correct_by_category += 1
+                end
+            end
+        end
+        @correct_by_category
+    end 
+
+    def percent_correct; (number_correct / @turns.count.to_f)*100 end
+
+    def percent_correct_by_category(input)
```
abdulredd, line 43: Try a more descriptive word like `category` here instead of `input` to lessen confusion. Using `input` makes it feel like it is input that is coming from user input.

The method works just fine, this is just for semantics.

lib/round.rb
```diff
@@ -0,0 +1,78 @@
...
+        counter = 0
+        card_count = deck.count
+        until turns.count == deck.count do
+            puts "This is card number #{counter += 1} out of #{card_count}."
+            puts "Question: #{current_card.question}"
+            guessing = gets.chomp
+            take_turn(guessing)
+            puts turns.last.feedback
+        end
+        if turns.count == deck.count
```
abdulredd, line 64: This `if` check isn't quite necessary since the `until` block that runs just before it runs until `turns.count == deck.count` so by the time that loop ends we can be sure that `turns.count == deck.count` so we don't need to check it again, and can just run the rest of the code inside the `if`

```ruby
until turns.count == deck.count do
  puts "This is card number #{counter += 1} out of #{card_count}."
  puts "Question: #{current_card.question}"
  guessing = gets.chomp
   take_turn(guessing)
  puts turns.last.feedback
end

puts "You had #{number_correct} correct guesses out of #{card_count} for a total score of #{percent_correct.to_i}%."
turns_correct = []
#[REST OF CODE...]
```

lib/turn.rb
```diff
@@ -0,0 +1,27 @@
...
+
+class Turn
+    attr_reader :card
+    attr_accessor :guess
+    def initialize(guess,card)
+        @guess = guess
+        @card = card
+    end
+
+    def correct?
```
abdulredd, line 11: Just wanted to suggest another way this method could have been written for brevity:

```ruby
def correct?
  @card.answer.capitalize == @guess.capitalize #<=== Conditionals equate to true or false by default so no need to do an explicit if/else
end
```

This is purely another way of writing this method, and I just wanted to mention it for reference.

lib/round.rb
```diff
@@ -0,0 +1,78 @@
+require 'pry'
+require_relative'./deck'
+require_relative'./turn'
+require_relative './card'
+
+class Round
+    attr_reader :deck, :turns, :correct_answer
```
abdulredd, line 7: There's an `attr_reader` here for `:correct_answer` but I don't see where it is ever being used.

PR #423, opened 2022-07-08
Astrid Hecht

Astrid-Hecht:


spec/card_generator_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+require './lib/turn.rb'
+require './lib/deck.rb'
+require './lib/round.rb'
+require './lib/card_generator.rb'
+
+RSpec.describe Round do
+    before :each do
+        @file = 'cards.txt'
+        @file_invalid = 'cards_invalid.txt'
+        @cardset = CardGenerator.new(@file)
```
abdulredd, line 11: The value that `@cardset` currently holds is and instance of `CardGenerator`, not the array of cards that were created in the the `CardGenerator` class. If you look at the the interaction pattern for iteration 4, you'll notice that after we pass a filename to `new` we are able to then call on a `cards` method on an instance of `CardGenerator` and get the array of cards returned. This is missing from your solution. 

I do believe this is what `temp_deck` is supposed to be, but just want to point out that the naming differences from the one provided in the interaction pattern.

lib/turn.rb
```diff
@@ -0,0 +1,27 @@
...
+class Turn
+  attr_reader :guess, 
+              :card
+
+  def initialize(guess, card) #create Turn
+    @card = card
+    @guess = guess
+  end
+
+  def correct? #return boolean value of if card guess matches the card's answer
```
abdulredd, line 12: The body of this method could have been written 2 other ways using less code, what ways do you believe those might be?

lib/turn.rb
```diff
@@ -0,0 +1,27 @@
...
+    if @guess.downcase == @card.answer.downcase
+      return true
+    else
+      return false
+    end
+  end
+
+  def feedback #return string version of above method
```
abdulredd, line 20: There's a little repetition here, there's code here that also shows up in the `correct?` method, we could have potentially used that method to help us here as well. This method could look something like this:

```ruby
def feedback
  if correct?
    return "Correct!"
  else
    return "Incorrect."
  end
end
```

How could this method be further refactored?

lib/turn.rb
```diff
@@ -0,0 +1,27 @@
+require './lib/card'
```
abdulredd, line 1: Because you are not creating `Card` objects from inside your `Turn` class, you do not need to `require` the `card.rb` file.

lib/deck.rb
```diff
@@ -0,0 +1,27 @@
+require './lib/card'
```
abdulredd, line 1: Because you are not creating `Card` objects from inside your `Deck` class, you do not need to `require` the `card.rb` file.

lib/round.rb
```diff
@@ -0,0 +1,41 @@
...
+    attr_reader :deck, :turns, :card_index
+     
+    def initialize(deck) #create Round
+        @deck = deck 
+        @turns = []
+        @card_index = 0 #set index to zero for future card cycling
+    end
+
+    def take_turn(guess) #store a guess as a Turn in an array and +1 index to move to next card
+        turn = Turn.new(guess, @deck.card_at(@card_index))
```
abdulredd, line 15: This piece of code, `@deck.card_at(@card_index)` is the same exact code that shows up in your `current_card` method, could you have replaced this line with a call to the `current_card` method instead instead of repeating the code?

lib/round.rb
```diff
@@ -0,0 +1,41 @@
...
+        @card_index += 1
+        @turns << turn 
+        return turn
+    end
+
+
+    def current_card #returns current card at the index value in the Deck
+        @deck.card_at(@card_index)
```
abdulredd, line 23: I like your use of a helper method here `card_at`. I also believe you could have used the number of turns in the `@turns` array to also determine what the current card should be since the number of turns in the `@turns` array sand the current value that `@card_index` holds will always be the same.

lib/round.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+
+    def number_correct #searches thru turn array for turns with correct guesses and counts them
+        @turns.select {|turn| turn.guess.downcase == turn.card.answer.downcase }.count
```
abdulredd, line 27: This line (`turn.guess.downcase == turn.card.answer.downcase`) has already been defined in your `Turn` classes `correct?` method and you can use that method here so you do not have to repeat the code like so:

```ruby
@turns.select {|turn| turn.correct? }.count
```

lib/round.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+
+    def number_correct_by_category(category) #same as above but adds matching category symbol to passed in symbol requirement
+        @turns.select {|turn| turn.guess.downcase == turn.card.answer.downcase && turn.card.category == category}.count
```
abdulredd, line 31:
```ruby
@turns.select {|turn| turn.correct? && turn.card.category == category}.count
```

lib/round.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+
+    def percent_correct #same as number correct but divides by total cards in deck (after converting both to floats), multiplies by 100 to get %, and then rounds to 1 decimal place
+        (((@turns.select {|turn| turn.guess.downcase == turn.card.answer.downcase }.count.to_f) / (@turns.count).to_f) * 100 ).round(1)
```
abdulredd, line 35: (((@turns.select {|turn| turn.correct? }.count.to_f) / (@turns.count).to_f) * 100 ).round(1)

lib/round.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+
+    def percent_correct_by_category(category) #same as percent_correct but adds category qualifier to the matching cards count
+        (((@turns.select {|turn| turn.guess.downcase == turn.card.answer.downcase && turn.card.category == category}.count.to_f) / (@turns.select {|turn| turn.card.category == category }.count).to_f) * 100 ).round(1)
```
abdulredd, line 39: This line is way too long and I would say the same thing for the line in `percent_correct`. While the ruby interpreter doesn't mind, it makes it hard for another developer to interpret the line easily.

My suggested would be to store the return value of each parenthesized equation into a variable and use those at the very end to do the math.

spec/card_spec.rb
```diff
@@ -1,27 +1,29 @@
...
-    expect(card.answer).to eq("Juneau")
+    expect(@card.answer).to be_a(String)
+    expect(@card.answer).to eq(@test_ans)
   end
 
   it 'has a category' do
-    card = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
-
-    expect(card.category).to eq(:Geography)
+    expect(@card.category).to be_a(Symbol)
```
abdulredd, line 26: Your explicitly stating that card categories should be symbols, but when they get imported from the `card.txt` files the card categories are strings, so this test, although passing, actually isn't correct. Even when you check the interaction pattern, it allows for categories that are not symbols.

spec/turn_spec.rb
```diff
@@ -0,0 +1,47 @@
+require './lib/card.rb'
+require './lib/turn.rb'
+require './lib/card_generator.rb'
+
+RSpec.describe Turn do
+  before :each do
```
abdulredd, line 6: It's super awesome that you are using a `before :each` to set up your tests to reduce repetition, but why do we need to create a `CardGenerator` object and `Deck` object just to be able to test the `Turn` class. The only class that should be included in the test setup in the `Card` class and the `Turn` class itself, all other classes don't need to be here. If we're following true TDD, these classes would be here during the testing of the `Turn` class because those classes are created **afterwards**.

spec/turn_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    @card = @deck.card_at(0)
+    @guess_wrong = "arbitrarily wrong"
+    @guess_correct = @card.answer
+    @turn_correct = Turn.new(@card.answer, @card)
+    @turn_wrong = Turn.new(@guess_wrong, @card)
+  end
+
+  it 'exists' do
+     expect(@turn_correct).to be_instance_of(Turn)
+     expect(@turn_wrong).to be_instance_of(Turn)
```
abdulredd, line 18: Not sure if we need to check if a correct turn **and** an incorrect turn are both `Turn` objects since we know that we need to create `Turn` objects in order to even pass a correct or incorrect guess.

spec/turn_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  it 'correctly created card within turn' do
+    expect(@card).to be_instance_of(Card)
```
abdulredd, line 22: Checking whether a card was an instance of the `Card` class was already handled in the `card_spec.rb` file and doesn't need to be verified again. 

In fact, I believe we can get rid of lines 22, 23, 24, 25 altogether because they all seem to be testing functionality of a`Card` object and not functionality of a `Turn` object.

spec/turn_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+  
+  it 'turns have correctly assigned guess' do 
+    expect(@turn_correct.guess).to be_a(String)
+    expect(@turn_wrong.guess).to be_a(String)
+    expect(@turn_correct.guess).to eq(@guess_correct)
+    expect(@turn_wrong.guess).to eq(@guess_wrong)
+  end
+
+
+  it 'identifies correct & incorrect' do
```
abdulredd, line 36: Great job testing for both sides of the coin, `true` as well as `false`.

spec/turn_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    expect(@turn_correct.correct?).to eq(true)
+    expect(@turn_wrong.correct?).to eq(false)
+  end
+  
+
+  it 'returns correct feedback' do
```
abdulredd, line 42: Great job testing for both sides of the coin, `"Correct!"` as well as `Incorrect.`.

spec/deck_spec.rb
```diff
@@ -0,0 +1,48 @@
+require './lib/card.rb'
+require './lib/deck.rb'
+require './lib/card_generator.rb'
+
+RSpec.describe Deck do
+
+  before :each do
+    @cardset = CardGenerator.new('cards.txt').temp_deck
```
abdulredd, line 8: `CardGenerator` shouldn't be a part of the `Deck` test.

spec/round_spec.rb
```diff
@@ -0,0 +1,129 @@
+require './lib/card.rb'
+require './lib/turn.rb'
+require './lib/deck.rb'
+require './lib/round.rb'
+require './lib/card_generator.rb'
+
+RSpec.describe Round do
+  before :each do
+    cardset = CardGenerator.new('cards.txt').temp_deck
```
abdulredd, line 9: Do we need this here?

spec/round_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+  it 'correctly takes a turn & stores argument' do
+    new_turn = @round.take_turn("Juneau")
+
+    expect(new_turn).to be_instance_of(Turn)
+    expect(new_turn.guess).to eq("Juneau")
+    expect(new_turn.card).to eq(@card)
+  end
+  
+
+  it 'identifies correct and incorrect guesses' do
```
abdulredd, line 48: These looping constructs in this test and in other tests up to this point isn't necessary and you should be limiting your tests to method calls on the object.

spec/card_generator_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+        @cardset.temp_deck.each do |card|
+            card_reader << "#{card.question},#{card.answer},#{card.category.to_s.tr(':','').tr('_',' ')}"
+        end
+        File.foreach(@file) do |line|
+            file_reader << line.tr("\n",'')
+        end
+        expect(card_reader).to eq(file_reader)
+    end
+
+    it 'catches invalid input' do
```
abdulredd, line 31: This tests interrupts the ability to run `rspec` in the command line and probably is not needed, looking at the code in the method.

PR #425, opened 2022-07-08
Dominic ODonnell v2

Dominicod:


spec/turn_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    expect(@turn.correct?).to eq(false)
+  end
+
+  it 'input in feedback is Incorrect' do
+    @turn = Turn.new('Dallas', @card)
+    @turn.correct?
+
+    expect(@turn.feedback).to eq('Incorrect.')
+  end
```
epintozzi, line 45: Nice work testing both outcomes for a correct (or incorrect) guess 🔥

lib/deck.rb
```diff
@@ -0,0 +1,20 @@
...
+  def count
+    @count = cards.count
+  end
+
+  def cards_in_category(category)
+    @categorized_cards = []
+    @cards.each do |card|
+      @categorized_cards << card if card.category == category
+    end
+  end
```
epintozzi, line 19: We want to be able to call `cards_in_category` and return a collection of cards. The way this method is written, it is returning `@cards` (that is was `each` returns and that is the last thing your method is doing). You've written it such that we have to call _another_ method (`categorized_cards`) to actually get this list, which is not what the spec is asking for. It also adds some unnecessary code.

spec/deck_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    expect(@deck.categorized_cards).to eq([@deck.cards[1], @deck.cards[2]])
+
+    @deck.cards_in_category(:Geography)
+
+    expect(@deck.categorized_cards).to eq([@deck.cards[0]])
+
+    @deck.cards_in_category('Pop Culture')
+
+    expect(@deck.categorized_cards).to eq([])
+  end
```
epintozzi, line 39: See comments on the `cards_in_category` method for more context, but this spec should be written like this in order to test what we are actually wanting this method to do:
```ruby
    expect(@deck.cards_in_category(:STEM)).to eq([@deck.cards[1], @deck.cards[2]])
    expect(@deck.cards_in_category(:Geography)).to eq([@deck.cards[0]])
    expect(@deck.cards_in_category('Pop Culture')).to eq([])
```

lib/round.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+class Round
+  attr_accessor :deck,
+                :turns,
+                :turn,
+                :current_card,
+                :number_correct,
+                :correct_by_category,
+                :percent_correct
```
epintozzi, line 10: There are an awful lot of `attr_readers` here. Typically we'd only have an `attr_reader` for 1) attributes that we initialize with and 2) that we need to access from outside the class. Some of these should be defined as their own methods below to follow ruby convention.

lib/round.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  def percent_correct_by_category(category)
+    correct_by_category = number_correct_by_category(category)
+    number_of_trys_in_category = 0
+
+    @turns.each do |turn|
+      number_of_trys_in_category += 1 if turn.card.category == category
+    end
+    (correct_by_category / number_of_trys_in_category) * 100
```
epintozzi, line 47: Since `correct_by_category` and `number_of_trys_in_category` are integers, if the result of the division is less than 1 it returns 0, which we then multiply by 100 and still returns 0. If I add a `.to_f` to `correct_by_category` the whole thing will convert to a float and it solves this issue.

spec/round_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    @new_turn = @round.take_turn('Venus')
+
+    expect(@round.turns.count).to eq(2)
+    expect(@round.turns.last.feedback).to eq('Incorrect.')
+    expect(@round.number_correct).to eq(1)
+  end
+  it 'percent and category checks' do
+    @new_turn = @round.take_turn('Juneau')
+    @new_turn = @round.take_turn('Venus')
+
```
epintozzi, line 73: If you update this test to include another turn for a correct STEM answer (ex: `@new_turn = @round.take_turn('North north west')`) and a corresponding assertion (`expect(@round.percent_correct_by_category(:STEM)).to eq(50.0)`) this will fail due to the issue noted above in the `Round` class on this method and the use of `Integer` instead of `Float`.

lib/round.rb
```diff
@@ -0,0 +1,49 @@
...
+    @correct_by_category
+  end
+
+  def percent_correct_by_category(category)
+    correct_by_category = number_correct_by_category(category)
+    number_of_trys_in_category = 0
+
+    @turns.each do |turn|
+      number_of_trys_in_category += 1 if turn.card.category == category
+    end
```
epintozzi, line 46: If you update your `cards_in_category` method on `Deck` to function as intended, you can use that here in place of this iterator to clean the code up a bit and make it less redundant.

spec/card_generator_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  it 'file is a file' do
+    expect(File.file?(@filename)).to eq(true)
+  end
+
+  it 'file is txt file' do
+    expect(File.extname(@filename)).to eq('.txt')
+  end
+
+  it 'cards == cards' do
+    expect(@cards).to eq(@cards)
```
epintozzi, line 30: A better test here might assert that `@cards` is an `Array` and that the array contains `Card` objects. Right now you are just testing that an object (`@cards`) is equal to itself (`@cards`) which will always be true regardless of the code in your `CardGenerator` class.

epintozzi: I would also probably check the size of the array. If you know `cards.txt` should've created 6 cards, we'd want to verify that as well.

lib/card_generator.rb
```diff
@@ -0,0 +1,31 @@
...
+    # file_data. Then it is indexing into the file, taking those values and putting them into
+    # my new variables
+    (0..file_data.count - 1).each do |num|
+      question = result[num][0]
+      answer = result[num][1]
+      category = result[num][2]
+      cards << (Card.new(question, answer, category))
+    end
+    # Automatically creates a new deck
+    Deck.new(cards)
```
epintozzi, line 29: I personally would just return the  collection of cards here. The class is `CardGenerator` not `DeckCreator`. This way, if there is another use case in the future for generating cards from a text file, we can re-use this class without having to remove the deck creation functionality. I'd just create the `Deck` in the runner with the cards that are returned from the generator.

PR #428, opened 2022-07-08
Andrew Mullins

mullinsand:


lib/turn.rb
```diff
@@ -0,0 +1,16 @@
...
+        @guess = guess
+        @card = card
+    end
+
+    def correct?
+        @guess == @card.answer
+    end
+
+    def feedback
+        return "Correct!" if correct? == true
```
abdulredd, line 13: Because of the way that booleans conditions work in programming you do not need to explicitly compare a boolean value in an `if` condition. Something like this also works.

```ruby
return "Correct!" if correct?
return "Incorrect." if !correct?
```

lib/deck.rb
```diff
@@ -0,0 +1,15 @@
+class Deck
+    attr_reader :cards, :card_categories
```
abdulredd, line 2: What is `card_categories`?

lib/flashcard_runner.rb
```diff
@@ -0,0 +1,38 @@
...
+def start
+    filename = "cards.txt"
+    cards = CardGenerator.new(filename).cards
+    deck = Deck.new(cards)
+    round = Round.new(deck)
+
+    puts "Welcome! You're playing with #{deck.count} cards."
+    puts "------------------------------------------------"
+
+    until round.number_of_rounds == deck.count do
```
abdulredd, line 17: This could this have been re-written where it's not using an extra instance variable `number_of_rounds` and instead keeps running until the number of cards in the deck is equal to the number of turns in the round.

There is also so improper indentation going on here and the code in the `until` block should probably be indented one over.

lib/card_generator.rb
```diff
@@ -0,0 +1,35 @@
...
+            line.split(",")
+        end
+    end
+
+    def open_file
+        @file_contents = File.read("./lib/"+@text_file)
+    end
+
+    def create_card(new_card, card_number, data_lines_array)
+        new_card = Card.new(data_lines_array[card_number][0].strip,data_lines_array[card_number][1].strip.downcase,data_lines_array[card_number][2].strip)
```
abdulredd, line 20: This line is a little too long and each part could actually be placed in a variable to be more easier to read.

spec/round_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+        it 'has attributes' do
+           
+            expect(@round.deck).to eq(@deck)
+            expect(@round.turns).to eq([])
+            expect(@round.current_card).to eq(@card_1)
+        end
+
+        it 'keeps track of current card' do
+
+            expect(@round.current_card).to eq(@card_1)
```
abdulredd, line 29: Isn't this expectation the exact same one on line 24? This test isn't needed.

spec/card_generator_spec.rb
```diff
@@ -0,0 +1,16 @@
+require './lib/card_generator'
+require './lib/card'
+filename = "cards.txt"
+describe CardGenerator do
```
abdulredd, line 4: Where is the test that checks for successful `CardGenerator` initialization?

PR #429, opened 2022-07-09
Ken Lenhart 

Penitent0:


Penitent0:
I was able to complete iteration 4 and successfully implement a card generator class that pulls questions from a text file and converts them into card classes. Unfortunately, I was not able to get my associated test for the card generator class to pass and I would like to discuss with an instructor to get some pointers. Thanks.

Penitent0:
I also realized that my flashcard_runner.rb file still has the absolute directory path listed for the cards text file that is being passed into the card generator class which I'm sure will break it's functionality. Will message instructors and ask if I can do another pull request to change it.

flashcard_runner.rb
```diff
@@ -0,0 +1,48 @@
+require './lib/card'
+require './lib/turn'
+require './lib/deck'
+require './lib/round'
+require './lib/card_generator'
+
+def start
```
epintozzi, line 7: I'm so glad you were able to write your runner to be dynamic enough to handle any number of cards or categories. Great work!

spec/turn_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+  end 
+  it 'returns a guess' do 
+    expect(@turn.guess).to eq("Sacramento")
+  end 
+  it 'returns correct?' do
+    expect(@turn.correct?).to eq(true)
+  end 
+  it 'returns feedback' do
+    expect(@turn.feedback).to eq("Correct!")
+  end
```
epintozzi, line 24: We should also have a version of this test that verifies behavior for incorrect answers.

lib/deck.rb
```diff
@@ -0,0 +1,14 @@
+class Deck 
+  attr_reader :cards
+  def initialize(cards)
+    @cards = cards
+  end 
+  def count 
+    cards.count
+  end 
+  def cards_in_category(category)
+    cards.select do |card|
```
epintozzi, line 10: Great choice of enumerable here with `select`!

lib/round.rb
```diff
@@ -0,0 +1,46 @@
+class Round
```
epintozzi, line 1: It's good practice to put an empty line in between each new method definition. It makes it a little easier to read.

lib/round.rb
```diff
@@ -0,0 +1,46 @@
...
+    @turn_count += 1
+    @turns << new_turn
+    if new_turn.correct?
+      @number_correct += 1 
+    end 
+    new_turn
+  end
+  def number_correct
+    @number_correct
+  end
```
epintozzi, line 28: This method is doing exactly the same thing as our `attr_reader` for `number_correct` so we actually can get rid of this method altogether.

lib/round.rb
```diff
@@ -0,0 +1,46 @@
...
+                :take_turn,    
+                :turn_count          
+  def initialize(deck)
+    @deck = deck
+    @turns = []
+    @number_correct = 0
+    @turn_count = 0
+  end 
+  def current_card
+    self.deck.cards[self.turns.count]
```
epintozzi, line 15: `self` is redundant here in both places. Ruby convention would say we should not include it.

lib/round.rb
```diff
@@ -0,0 +1,46 @@
+class Round 
+  attr_accessor :deck, 
+                :turns, 
+                :current_card,
```
epintozzi, line 4: `current_card` is a method you are defining on line 14, so we do not also want to have an `attr_reader` for it. It can lead to conflicts to have methods by the same name. In this case, I would remove it as an `attr_reader` altogether.

lib/round.rb
```diff
@@ -0,0 +1,46 @@
+class Round 
+  attr_accessor :deck, 
+                :turns, 
+                :current_card, 
+                :number_correct, 
+                :take_turn,
```
epintozzi, line 6: Like `current_card`, `take_turn` is a method you are defining elsewhere in the class so we should remove it from our `attr_readers` to avoid redundancy and conflict.

lib/round.rb
```diff
@@ -0,0 +1,46 @@
...
+  def number_correct_by_category(correct_category)
+    correct_by_category = 0
+    self.turns.each do |turn|
+      if turn.card.category == correct_category 
+        if turn.correct?
+          correct_by_category += 1
+        end 
+      end 
+    end 
+    correct_by_category
```
epintozzi, line 38: We can clean this method up a lot by using the build in `count` method. Something like this:
```ruby
    turns.count do |turn|
      turn.correct? && turn.card.category == correct_category
    end
```

PR #430, opened 2022-07-09
Alaina Kneiling

Alaina-Noel:


spec/round_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+    card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    card_2 = Card.new("The Viking spacecraft sent back to Earth photographs and reports about the surface of which planet?", "Mars", :STEM)
+    card_3 = Card.new("Describe in words the exact direction that is 697.5° clockwise from due north?", "North north west", :STEM)
+    cards = [card_1, card_2, card_3]
+    deck = Deck.new(cards)
+    round = Round.new(deck)
+    round.take_turn("Juneau")
+    round.take_turn("Venus")
+
+    expect(round.percent_correct_by_category(:Geography)).to eq(50.0)
```
abdulredd, line 121:
```
expected: 50.0
got: 100.0
```

I want to go out on a limb and say that the expectation for this test is incorrect. If you look at the setup, there are 3 cards, 1 is a Geography card, 2 are STEM cards. We answer the Geography question correctly on line 118, and your test is checking the percentage correct by category and passing `:Geography` which we got right, so the expectation you actually be `eq(100.0)` and not `eq(50.0)`

Alaina-Noel: Yep! I see that now. You are correct.

lib/round.rb
```diff
@@ -0,0 +1,66 @@
...
+  def percent_correct_by_category(category)
+    ((correct_turns.select { |turn| turn.card.category == category}.size).to_f/
+    (((turns.select { |turn| turn.card.category == category}.size)))*100).round(1)
+  end
+
+  def start
+    puts "Welcome! You're playing with #{deck.cards_as_array.size} cards.
+        -------------------------------------------------
+        This is card number 1 out of #{deck.cards_as_array.size}.
+        puts Question: #{current_card.question}"
```
abdulredd, line 45: There is a "puts" word here that also gets printed out the the screen for the first card.

lib/round.rb
```diff
@@ -0,0 +1,66 @@
...
+      if turns.size < deck.cards_as_array.size
+        puts "This is card number #{(turns.size + 1)} out of #{deck.cards_as_array.size}."
+      end
+    if turns.size == deck.cards_as_array.size
+      puts "****** Game over! ******
+        You had #{correct_turns.size} correct guesses out of #{deck.cards_as_array.size} for a total score of #{percent_correct}%.
+        Mod1 Review Questions - #{percent_correct_by_category(":CS")}% correct
+        Geography - #{percent_correct_by_category(":Geography")}% correct"
+    else
+      puts "Question #{current_card.question}"
```
abdulredd, line 60: Small little formatting issue here but there is a colon (:) missing between the word "Question" and the cards question.

lib/cards.txt
```diff
@@ -0,0 +1,7 @@
+What is the capital of Alaska?,Juneau,:Geography
+In what state is Turing located?,Colorado,:Geography
+The extra file we can make to help us test & run code is called the ________ file. ,  runner , :CS
```
abdulredd, line 3: For this third card, there is are extra spaces at the beginning and end of "  runner ", therefore if you just enter an answer of "runner" it will come back as incorrect unless you include the extra spaces at the beginning and end of the word.

lib/round.rb
```diff
@@ -0,0 +1,66 @@
...
+  def show_next_card(user_guess)
+      take_turn(user_guess)
+      puts "#{turns[-1].feedback}"
+      if turns.size < deck.cards_as_array.size
+        puts "This is card number #{(turns.size + 1)} out of #{deck.cards_as_array.size}."
+      end
+    if turns.size == deck.cards_as_array.size
+      puts "****** Game over! ******
+        You had #{correct_turns.size} correct guesses out of #{deck.cards_as_array.size} for a total score of #{percent_correct}%.
+        Mod1 Review Questions - #{percent_correct_by_category(":CS")}% correct
```
abdulredd, line 57: Your solution should be able to grab the category of cards from the the deck and use that in the output. Right now, your categories are hard-coded and therefore aren't dynamic. The categories of the cards in your `cards.txt` file are Geography and CS, yet the categories that print out at the completion of the game are Geography and Mod1 Review Questions.

We should always try and code our solutions to be dynamic and work regardless of what the program receives.

lib/round.rb
```diff
@@ -0,0 +1,66 @@
+class Round
+  attr_accessor :deck,
+                :current_card,
+                :turns,
+                :correct_turns
```
abdulredd, line 5: You only ever use `correct_turns` inside the `Round` class, so you do not need to add it to the `attr` list.

lib/card_generator.rb
```diff
@@ -0,0 +1,23 @@
+#Cannot require a txt file in ruby up here
+require 'pry'
+require './lib/card.rb'
+
+class CardGenerator
+
+    attr_reader :txt_file, :lines, :cards
```
abdulredd, line 7: We could probably do without creating readers for `txt_file` and `lines` and the only variable we need to have access to is `cards` so it should be the only one in this list.

lib/turn.rb
```diff
@@ -0,0 +1,20 @@
...
+    @guess = guess
+    @card = card
+  end
+
+  def correct?
+    @card.answer == @guess
+  end
+
+  def feedback
+    correct? ? "Correct!" : "Incorrect."
```
abdulredd, line 16: Great user of the ternary operator here.

lib/turn.rb
```diff
@@ -0,0 +1,20 @@
...
+class Turn
+  attr_accessor :card, :guess
+
+  def initialize(guess, card)
+    @guess = guess
+    @card = card
+  end
+
+  def correct?
+    @card.answer == @guess
```
abdulredd, line 12: This is great, the only thing missing it allowing for mismatching case.

PR #431, opened 2022-07-09
Ryan Smith

RyanChrisSmith:
Through to end of iteration 3, not finished


RyanChrisSmith:
Updating to the end of iteration 4

cards.txt
```diff
@@ -0,0 +1,7 @@
+What is 5 + 5?,10,STEM
+What is Rachel's favorite animal?,red panda,Turing Staff
+What is Mike's middle name?,nobody knows,Turing Staff
+What cardboard cutout lives at Turing?,Justin bieber,PopCulture
+What is the pleasant smell that frequently accompanies the first rain after a long period of warm dry weather called?,petrichor,Vocabulary
+How many hours did Ryan spend in week 1 on flash cards?,Infinite,Random
```
epintozzi, line 6: Infinity

flashcard_runner.rb
```diff
@@ -0,0 +1,21 @@
+require './lib/card'
+require './lib/turn'
+require './lib/deck'
+require './lib/round'
+require './lib/card_generator'
+ 
+cards = CardGenerator.new("cards.txt").cards
+deck = Deck.new(cards)
+round = Round.new(deck)
+round.start
```
epintozzi, line 10: 🔥 This is great. Very concise!

flashcard_runner.rb
```diff
@@ -0,0 +1,21 @@
...
+    # def setup 
+    #     # @card_1 = Card.new("What is 5 + 5?", "10", :Mathematics)
+    #     # @card_2 = Card.new("What is Rachel's favorite animal?", "red panda", "Turing Staff")
+    #     # @card_3 = Card.new("What is Mike's middle name?", "nobody knows", "Turing Staff")
+    #     # @card_4 = Card.new("What cardboard cutout lives at Turing?", "Justin Bieber", "Pop Culture")
+    #     @cards = CardGenerator.new("cards.txt")
+    #     require 'pry' ; binding.pry
+    #     @deck = Deck.new(@cards)
+    #     @round = Round.new(@deck)
+    # end
```
epintozzi, line 21: Generally it's considered best practice to cleanup unused or commented out code before submitting a PR/merging to main.

lib/turn.rb
```diff
@@ -0,0 +1,22 @@
+class Turn
+attr_reader :guess, 
+            :card, 
+            :feedback
```
epintozzi, line 4: We don't need the `feedback` `attr_reader` because this method is also defined below on line 15.

lib/turn.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def correct?
+    @guess == card.answer
+  end
+
+  def feedback
+    if @guess == card.answer
+      'Correct!'
+    else 'Incorrect.'
+    end
```
epintozzi, line 19: While this works, ruby convention would have us put `'Incorrect.'` on its own line like this:
```ruby
...
else
  'Incorrect.'
end
```

spec/turn_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    @turn = Turn.new("Denver", @card)
+    expect(@turn.correct?).to eq(false)
+  end 
+
+  it 'gives feedback' do
+    expect(@turn.feedback).to eq('Correct!')
+
+    @turn = Turn.new("Saturn", @card)
+    expect(@turn.feedback).to eq('Incorrect.')
+  end
```
epintozzi, line 33: Great job testing both outputs (correct and incorrect). We always want to test every possible outcome.

lib/deck.rb
```diff
@@ -0,0 +1,19 @@
...
+    def initialize(cards) 
+        @cards = cards
+    end 
+
+    def count
+        @cards.count 
+    end 
+
+    def cards_in_category(specific_category)
+        @cards.select{|card| card.category == specific_category}
```
epintozzi, line 13: Great use of `select` here!

lib/deck.rb
```diff
@@ -0,0 +1,19 @@
...
+    end   
+    
+    def categories
```
epintozzi, line 16: 🔥 I haven't looked at your `Round` class yet but I have a strong hunch about what this method is used for and I'm very happy to see it. This allows your program to be very dynamic which is pretty much always a good thing.

spec/round_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+        @round.take_turn("Venus")
+
+        expect(@round.percent_correct).to eq(50.0)        
+    end 
+
+    it 'can show percentage correct by category' do
+        new_turn = @round.take_turn("Juneau")
+        @round.take_turn("Venus")
+        
+        expect(@round.percent_correct_by_category(:Geography)).to eq(100.0)
```
epintozzi, line 79: You have this set up (with the `take_turn("Venus")`) to do another assertion for percentage less than 100% so I would add something like this to verify you aren't getting a false positive with the one existing assertion.
`expect(@round.percent_correct_by_category(:STEM)).to eq(0.0)`

In fact I'd probably add a 3rd turn with a correct STEM answer to verify it can calculate 50% (or something other than 100 or 0).

lib/round.rb
```diff
@@ -0,0 +1,77 @@
...
+    end 
+
+    def number_correct_by_category(category)
+        @correct_count = 0
+        @turns.each do |turn|
+            if turn.card.category == category && turn.correct? == true
+                @correct_count += 1
+            end 
+        end 
+        @correct_count
```
epintozzi, line 36: Try to see if you can refactor this using a different enumerable method. `count` might be a nice one here.

lib/round.rb
```diff
@@ -0,0 +1,77 @@
...
+        (number_correct / turns.count.to_f) * 100
+    end 
+
+    def percent_correct_by_category(category)
+        total_correct_by_category = 0
+        turns.each do |turn|
+            if turn.card.category == category 
+                total_correct_by_category +=1
+            end 
+        end
```
epintozzi, line 49: Your `cards_in_category` method in your `Deck` class already pretty much does exactly this. Can you think of a way to refactor to use that method here?

lib/round.rb
```diff
@@ -0,0 +1,77 @@
...
+            if turn.card.category == category 
+                total_correct_by_category +=1
+            end 
+        end 
+        (number_correct_by_category(category) / total_correct_by_category.to_f)*100 
+    end     
+
+    def categories
+        @deck.categories        
+    end
```
epintozzi, line 55: There's nothing wrong with having this method here, but I would probably just call `@deck.categories` directly on line 73 instead of having an extra method for it.

lib/round.rb
```diff
@@ -0,0 +1,77 @@
...
+    def categories
+        @deck.categories        
+    end 
+    
+    def start
+        puts "Welcome! You're playing with #{@deck.cards.count} cards.\n-------------------------------------------------"
+        
+        tally = 0
+        card_count = @deck.count 
+        until @turns.count == @deck.count do
```
epintozzi, line 62: I think you could use the `card_count` variable defined above here, unless I'm missing something.

```suggestion
        until @turns.count == card_count do
```

lib/round.rb
```diff
@@ -0,0 +1,77 @@
...
+            take_turn(guess)
+            puts turns.last.feedback
+        end 
+        
+        puts "****** Game over! ******"
+        puts "You had #{number_correct} correct guesses out of #{card_count} for a total score of #{percent_correct.to_i}%."
+        
+        categories.each do |category|     
+            puts  "#{category} - #{percent_correct_by_category(category).to_i}% correct"
+        end
```
epintozzi, line 75: This is excellent. I love that you made this dynamic so it can change with every deck regardless of what past or future categories we might have.

spec/card_generator_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+        @card_generator = CardGenerator.new(@filename)
+        @cards = @card_generator.cards
+    end 
+
+    it 'exists' do             
+        expect(@card_generator).to be_instance_of(CardGenerator)
+    end 
+    
+    it 'reads file' do
+        expect(@card_generator.cards).to eq(@cards)
```
epintozzi, line 16: I would also test:
```ruby
expect(@card_generator.cards).to be_a(Array)
expect(@card_generator.cards.count).to eq(6) #or however many it is
```

PR #432, opened 2022-07-09
Sergio Azcona

Sergio-Azcona:


lib/card.rb
```diff
@@ -0,0 +1,11 @@
...
+class Card
+    attr_reader :question, :answer, :category
+    def initialize (question, answer, category)
+        @question = question
+        @answer = answer
+        @category = category
+
+    end
+
+end
```

lib/deck.rb
```diff
@@ -0,0 +1,15 @@
...
+        @cards = cards     
+    end
+
+    def count
+        @cards.length 
+    end
+
+    def cards_in_category(category)
+       #need to select cards where card cateogry == category
+        @cards.select{|card| category == card.category}
```
megstang, line 13: great use of `select` here

lib/round.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def current_card
+    @deck.cards[0]
+  end
+
+  def take_turn(guess)
+    new_turn = Turn.new(guess, current_card)
+    @turns << new_turn
+    new_turn
+  end
```
megstang, line 18: When you take a turn, you also move to the next card, so it would be good to add some logic in this method to move the top card off of the top of the array so that the next card becomes the 'current card'.

lib/turn.rb
```diff
@@ -0,0 +1,26 @@
+class Turn
+    attr_reader :guess, :card #these will recieve input; if only output do not include
+    def initialize(guess, card)#arguments 
+        @guess = guess
+        @card = card
+    end
+    
+    def correct?
+        @guess == card.answer
```
megstang, line 9: In this method, you're calling to card instead of @card. This works, because you're calling to the attr_reader on line 2, which returns the instance variable on line 5. Since you have access to the instance variable, call to that rather than the attr_reader. The purpose of the attr reader is so that you can call to that attribute when you're outside of this class.

spec/deck_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+
+    it 'should return the count of card' do
+        expect(@deck.count).to eq(3)
+    end
+
+    it 'should return the category' do
+        expect(@deck.cards_in_category(:STEM)).to eq([@card_2, @card_3])
+    end
+
```
megstang, line 29: This test looks great! The only thing I'll add, is that it might be nice to add more assertions to the last test where you're testing `cards_in_category`. For example, maybe you make another assertion like: 
`expect(@deck.cards_in_category(:Geography)).to eq([@card_1])`

This will help make sure that your code is dynamic and working for any category that gets passed in to the method

spec/round_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+RSpec.describe Round do
+  before :each do
+    @card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    @card_2 = Card.new("The Viking spacecraft sent back to Earth photographs and reports about the surface of which planet?", "Mars", :STEM)
+    @card_3 = Card.new("Describe in words the exact direction that is 697.5° clockwise from due north?", "North north west", :STEM)
+    @cards = [@card_1, @card_2, @card_3]
+    @deck = Deck.new(@cards) 
+    @round = Round.new(@deck)
+    @turn = Turn.new("Juneau", @card)
```
megstang, line 16: You do not need this in this test.

spec/round_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  it 'should return new turn array' do
+    new_turn = @round.take_turn("Juneau")
+    expect(@round.turns).to eq([new_turn])
+  end
+
+  it 'should give current number'
+    expect(@round.number_correct).to eq(1)
+  end
```
megstang, line 47: This will not pass because `@round` is a new instance of the round class that has yet to have a turn taken. So, number_correct should be 0 once the round starts brand new.

megstang: If you wanted to test that this method worked after a turn was taken, we could do something like this

```ruby
  it 'should give current number'
    @round.take_turn("Juneau")
    expect(@round.number_correct).to eq(1)
  end
```

spec/turn_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+        expect(@turn.card).to eq(@card) 
+    end
+
+    it 'should return correct?' do#
+        expect(@turn.correct?).to eq(true)
+    end
+
+    it 'should return feedback' do
+        expect(@turn.feedback).to eq('Correct!') 
+    end
```
megstang, line 37: We'll also need to test the instance of when the turn object is incorrect.

megstang (commented):
@Sergio-Azcona I left some comment on your PR. Please let me know if you have any questions. I did not see a submission come through on the submission doc that was linked in the calendar due date event. In the future, please make sure to complete those submission forms, becuase it allows you to give us a bit more information about your process.

megstang:
I'll follow up later today with your scores via slack.

Sergio-Azcona:
thank you!! i thought i submitted the submission... but i cannot be 100% sure.
noted for next time though.
thank you!!

PR #433, opened 2022-07-09
Gavin Carew

gjcarew:


spec/card_generator_spec.rb
```diff
@@ -0,0 +1,10 @@
+require './lib/card_generator.rb'
+# require './lib/cards.txt'
+RSpec.describe CardGenerator do
+  it 'exists' do
+    new_cards = CardGenerator.new('./lib/cards.txt')
```
abdulredd, line 5: This test currently fails, and I see you have a `pry` in here to try and debug it. The only reason this is failing is because you are not actually returning the array of cards but the actual `CardGenerator` object itself. Line 5 would actually have to be: `new_cards = CardGenerator.new('./lib/cards.txt').cards` to return the cards array and then the test passes.

lib/cards.txt
```diff
@@ -0,0 +1,8 @@
+What is the fastest living creature,Mantis shrimp,Zoology
```
abdulredd, line 1: This is being a little nit-picky I'll admit, but your questions are all missing question marks.

lib/round.rb
```diff
@@ -0,0 +1,91 @@
...
+
+  def start
+    puts "Welcome! You're playing with #{@deck.count} cards"
+    puts "_______________________________________________________"
+
+    @deck.count.times do
+      puts "This is card number #{@turns.count + 1} of #{@deck.count}"
+      puts "Question: #{current_card.question}"
+
+      guess = gets.chomp
```
abdulredd, line 78: Whenever you are taking response from and end user, you should always try to implement some feature that disregards casing. Maybe `downcase` or `upcase` the users response and compare that to a downcased or upcased version of the cards answer.

lib/deck.rb
```diff
@@ -0,0 +1,19 @@
+require './lib/card'
+require './lib/deck'
+
+class Deck
+  attr_reader :cards
+  def initialize(cards = [])
```
abdulredd, line 6: Not sure the reason behind setting a default argument for cards and setting it to and empty array, we know that every time we instantiate a new `Deck` we will be passing and array of cards so you don't need this default argument.

lib/deck.rb
```diff
@@ -0,0 +1,19 @@
...
+    @cards = cards
+  end
+# QUESTION: Should I overwrite the default count method?
+  def count
+    @cards.count
+  end
+
+  def cards_in_category(category)
+    @cards.select do |card|
```
abdulredd, line 15: Great use of the `select` method here and trying out other enumerable's besides `each`.

lib/turn.rb
```diff
@@ -0,0 +1,19 @@
...
+    @guess = guess
+    @card = card
+  end
+
+  def correct?
+    @guess == @card.answer
+  end
+
+  def feedback
+    return "Correct!" if guess == card.answer
```
abdulredd, line 16: Awesome use of using explicit and implicit returns to your advantage!

lib/round.rb
```diff
@@ -0,0 +1,91 @@
...
+  #
+  def current_card
+    @deck.cards[@turns.count]
+  end
+
+  def take_turn(guess)
+    turn = Turn.new(current_card, guess)
+    @turns << turn
+
+    if turn.correct? == true
```
abdulredd, line 24: Because `correct?` will either return `true` or `false`, you don't need to compare it to `true` or `false` again... so this line could be re-written like:

```ruby
if turn.correct?
  @correct_turns << turn
end
```

lib/round.rb
```diff
@@ -0,0 +1,91 @@
...
+    end
+  end
+
+
+  def percent_correct
+    (@correct_turns.count.to_f/@turns.count.to_f) * 100
+  end
+
+  def percent_correct_by_category(category)
+    correct = @correct_turns.count do |turn|
```
abdulredd, line 46: I know this code was probably indented like this to help you read it better but unfortunately this is improper indentation and should try and stick to ruby conventions as closely as possible.

spec/deck_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    @cards = [@card_1, @card_2, @card_3]
+    @deck = Deck.new(@cards)
+  end
+
+  it 'exists' do
+    expect(@deck).to be_instance_of(Deck)
+  end
+
+  it 'contains cards' do
+    expect(@deck.cards).to be(@cards)
```
abdulredd, line 27: I like the use of `be` here. `be` and `eq` behave in 2 entirely different ways and it looks like you understand what those differences are.

spec/deck_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  it 'can count cards in a deck' do
+    expect(@deck.count).to eq(3)
+  end
+
+  it 'has a method to find cards in a category' do
+    expect(@deck.cards_in_category(:STEM)).to eq([@card_2, @card_3])
```
abdulredd, line 35: I would have added additional tests here to coincide with the interaction pattern to check if you can get the cards of the other category, and even of a category that doesn't exist.

spec/round_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+    @round = Round.new(@deck)
+
+  end
+
+  it 'exists' do
+    expect(@round).to be_instance_of(Round)
+  end
+
+  it 'keeps track of turns' do
+    expect(@round.turns).to be_empty
```
abdulredd, line 33: `be_empty` ❤️

PR #434, opened 2022-07-09
Madeline Mauser Flash Cards

MadelineMauser:
Submission for the Flash Card project.

### DIFF
--- a/flashcard_runner.rb
+++ b/flashcard_runner.rb
@@ -1,0 +1,4 @@
+require './lib/card'
+require './lib/turn'
+require './lib/deck'
+require './lib/round'
--- a/flashcard_runner.rb
+++ b/flashcard_runner.rb
@@ -31,6 +31,7 @@ def start
   puts "****** Game over! ******"
   puts "You had #{@round.number_correct} correct guesses out of #{@round.deck.cards.length} for a total score of #{@round.percent_correct.round(0)}%."
   puts "Biology - #{@round.percent_correct_by_category(:Biology).round(0)}% correct"
+  puts "Chemistry - #{@round.percent_correct_by_category(:Chemistry).round(0)}% correct"
 
 end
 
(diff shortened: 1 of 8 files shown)


lib/turn.rb
```diff
@@ -0,0 +1,23 @@
...
+
+# Checks whether the guess matches the answer
+  def correct?
+     @guess == @card.answer
+  end
+
+# Prints feedback depending on whether guess was correct
+  def feedback
+    if @guess == @card.answer
+      puts "Correct!"
```
abdulredd, line 18: For this method you don't actually want to `puts` "Correct" or "Incorrect" but instead just return those values and change lines 57 and 63 in your `Round` classes `start` method to `puts` the feedback there.

spec/deck_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    @deck = Deck.new(@cards)
+  end
+
+
+  it 'exists' do
+    expect(@deck).to be_instance_of(Deck)
+  end
+
+  it 'has an array of @cards' do
+    expect(@deck.cards).to match_array([@card_1, @card_2, @card_3])
```
abdulredd, line 22: Great use of the `match_array` matcher here!

spec/round_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  it 'stores an array of turn instances' do
+    expect(@round.turns).to eq([])
+  end
+
+  it 'has a current card for a turn' do
+    expect(@round.current_card).to eq(@card_1)
+  end
+
+  it 'can take a new turn' do
```
abdulredd, line 36: I love this "can take a new turn" and the next "can take a second turn" where you have several expectations for each. Really robust testing!

flashcard_runner.rb
```diff
@@ -0,0 +1,19 @@
+require './lib/card'
+require './lib/turn'
+require './lib/deck'
+require './lib/round'
+
+@card_1 = Card.new("Broccoli, cabbage, and brussel sprouts are all the same species, True or False?", "True", :Biology)
```
abdulredd, line 6: Is there a reason why all these variable in your runner file are all instance variables and not just normal variables?

PR #435, opened 2022-07-09
Mary Ballantyne

mballantyne3:


flashcard_runner.rb
```diff
@@ -0,0 +1,52 @@
...
+        end
+    end
+
+    categories.each do |category|
+        puts "#{category} - #{round.percent_correct_by_category(category).to_i}% correct"
+    end
+
+end
+
+start
```
megstang, line 51: At this point in your developer career, I'm happy with what you were able to produce, and how you were able to get your game to run via this runner file. In the future, we'll challenge you to store as little code as possible outside of a class. Right now, you have code in this file that is not in a class. So, if you were to follow that convention, you could have a line of code in your runner file that was something like Game.start. Where Game is a new class that you created for the purpose of prepping the objects, and running the game. And start was a method that kicked off the flow of the game. The idea is that the flow of the game is stored in methods in that game class.

lib/card.rb
```diff
@@ -0,0 +1,9 @@
+class Card
+    attr_reader :question, :answer, :category
+    def initialize(question, answer, category)
+        @question = question
+        @answer = answer
+        @category = category
+    end
+end
+
```

lib/card_generator.rb
```diff
@@ -0,0 +1,18 @@
...
+
+    def initialize(filename)
+        lines = File.read(filename).strip.lines
+        @cards = lines.map do |line|
+            question, answer, category = line.strip.split(",")
+            Card.new(question, answer, category)
+        end
+    end
+
+end
```
megstang, line 18: Really great job here!

lib/deck.rb
```diff
@@ -0,0 +1,17 @@
+class Deck 
+    attr_reader :cards
+    
+    def initialize(cards) 
+        @cards = cards
+    end
+
+    def count
+        cards.count
+    end
```
megstang, line 10: In this method, you're calling to `cards` instead of `@cards`. This works, because you're calling to the attr_reader on line 2, which returns the instance variable on line 5. Since you have access to the instance variable, call to that rather than the attr_reader. The purpose of the attr reader is so that you can call to that attribute when you're outside of this class.

lib/deck.rb
```diff
@@ -0,0 +1,17 @@
...
+
+    def cards_in_category(category)
+        cards.select do |card|
+            card.category == category 
+        end
```
megstang, line 15: nice use of select 👍

lib/round.rb
```diff
@@ -0,0 +1,41 @@
...
+        @turns << new_turn
+        @deck.cards.rotate!
+        new_turn 
+    end
+
+    def number_correct
+        @turns.count do |turn|
+            turn.correct?
+        end
+    end
```
megstang, line 23: nice use of `count`

lib/round.rb
```diff
@@ -0,0 +1,41 @@
...
+    def percent_correct
+        number_correct.to_f / @turns.count.to_f * 100.0
+    end
+
+    def percent_correct_by_category(category)
+        total_turns_by_category = @turns.count do |turn|
+            turn.card.category == category 
+        end
+        number_correct_by_category(category).to_f / total_turns_by_category.to_f * 100.0
+    end
```
megstang, line 40: really great work on this one. Syntax looks good!

lib/turn.rb
```diff
@@ -0,0 +1,17 @@
...
+    def correct?
+        @guess == card.answer
+    end
+
+    def feedback
+        if correct? == true then "Correct!"
+        else "Incorrect"
+        end
+    end
+end
```
megstang, line 17: 👍

spec/turn_spec.rb
```diff
@@ -0,0 +1,51 @@
+require './lib/turn'
+require './lib/card'
+require 'rspec'
+
+RSpec.describe Card do 
+    it 'initializes a card' do
+        card = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+        expect(card).to be_instance_of(Card)
+    end
+end
```
megstang, line 10: Even though our Turn class relies on getting a card object passed to it, we do not need to test the creation of card objects in this spec, because we already have tested that a card can be created and has all of its attributes in the card_spec. So, for that reason, you can get rid of this describe block.

spec/turn_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+        card = Card.new("Which planet is closest to the sun?", "Mercury", :STEM)
+        turn = Turn.new("Saturn", card)
+        expect(turn.correct?).to eq(false)
+    end
+    
+    it 'gives negative feedback' do
+        card = Card.new("Which planet is closest to the sun?", "Mercury", :STEM)
+        turn = Turn.new("Saturn", card)
+        expect(turn.feedback).to eq("Incorrect")
+    end
```
megstang, line 47: These tests all look very good. I know you requested feedback on this spec, but each of these tests look good, and the creation of the objects that you have in each test make a lot of sense.

spec/round_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+        expect(new_turn.correct?).to eq(false)
+        expect(round.turns.count).to eq(2)
+        expect(round.turns.last.feedback).to eq("Incorrect")
+        expect(round.number_correct).to eq(1)
+        expect(round.number_correct_by_category(:Geography)).to eq(1)
+        expect(round.number_correct_by_category(:STEM)).to eq(0)
+        expect(round.percent_correct).to eq(50.0)
+        expect(round.percent_correct_by_category(:Geography)).to eq(100.0)
+        expect(round.current_card).to eq(card_3)
+    end
```
megstang, line 57: You're right, this is a large test. I actually am fine with it. This take_turn method was a big one, and it affected multiple pieces of your code. In order to be sure that all of the affected areas are working appropriately, I think your thoroughness is great. Some could argue that you don't need to test the methods on the new_turn object since those methods have already been tested in the Turn spec, but I like that you're testing them. By testing the turn object, you've created an integration test, which means that you are testing all classes that are involved in that one method.

megstang (commented):
@mballantyne3 Great job! I'll follow up in slack with your scores later today. Please let me know if you ahve any questions on my comments.

PR #438, opened 2022-07-10
A.J. Krumholz

ajkrumholz:


flashcard_runner.rb
```diff
@@ -0,0 +1,21 @@
...
+
+# prompt user for total number of cards in the game, scaleable
+print "How many cards would you like to play with? (1-#{cards.count}): "
+custom_deck_size = gets.chomp.to_i
+cards = cards.sample(custom_deck_size)
+
+deck = Deck.new(cards)
+round = Round.new(deck)
+
+round.start
```
megstang, line 21: Runner file looks good! I know we talked a bit about this at the check in. Another way to push yourself to slim this up a bit is to create some sort of CardGenerator class that can take care of the logic on lines 3-11

lib/card.rb
```diff
@@ -0,0 +1,9 @@
+class Card
+  attr_reader :question, :answer, :category
+
+  def initialize(question, answer, category)
+    @question = question
+    @answer = answer
+    @category = category
+  end
+end
```

lib/deck.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def count
+    cards.count
+  end
+
+  def cards_in_category(cat)
+    cards.select do |card|
+      card.category == cat
+    end
```
megstang, line 17: great use of `select`

lib/flashcard_generator.rb
```diff
@@ -0,0 +1,13 @@
...
+    puts "Welcome to Flashcard Generator\n\n\n"
+    puts "Each flashcard needs 1) a question, 2) an answer, 3) a category.\n\n\n"
+    print "Enter your question: "
+    question = gets.chomp
+    print "\n\nEnter the answer: "
+    answer = gets.chomp.split.map(&:capitalize).join(' ')
+    print "\n\nFinally, enter the question category: "
+    category = gets.chomp.capitalize
+    file << "\n#{question}/#{answer}/#{category}"
+end
```
megstang, line 13: Are you using this anywhere? I couldn't see it, but maybe I missed it.

lib/round.rb
```diff
@@ -0,0 +1,96 @@
+require './lib/deck'
+
+class Round
+  attr_reader :deck,
+              :turns
+  attr_accessor :correct_answers,
+                :turn_count,
+                :total_cards,
+                :categories
```
megstang, line 9: It doesn't look like you're trying to update these values outside of this class, so for that reason, these do not need to be attr_accessors.

lib/round.rb
```diff
@@ -0,0 +1,96 @@
...
+    @turn_count += 1
+    new_turn = Turn.new(answer, card)
+    @turns << new_turn
+    @correct_answers += 1 if @turns.last.correct?
+    new_turn
+  end
+
+  def number_correct
+    @correct_answers
+  end
```
megstang, line 35: you could change the instance variable to be called `@number_correct`, and then keep an attr_reader for it so you can read it from other classes.

megstang: ^^ if you did that, you wouldn't need this method

lib/round.rb
```diff
@@ -0,0 +1,96 @@
...
+
+  def percent_correct
+    (@correct_answers.to_f / turns.count) * 100
+  end
+
+  def number_correct_by_category(cat)
+    num_by_cat = turns.count do |turn|
+      turn.correct? && turn.card.category == cat
+    end
+  end
```
megstang, line 45: Nice work here. You don't necessarily need to save the return value of the enumerable into `num_by_cat`, because it will automatically get returned from this method.

lib/round.rb
```diff
@@ -0,0 +1,96 @@
...
+      turn.card.category == cat
+    end
+    (number_correct_by_category(cat).to_f / total_by_cat) * 100
+  end
+
+  # this seems like an ideal place to use #select, but
+  # @categories = deck.cards.select do |card|
+  #   card.category if @categories.include?(card.category)
+  # end
+  # doesn't pass tests for method
```
megstang, line 58: I believe that this won't work because `select` only returns the elements of the array that return true for the condition inside the block. I'm not sure that you can select something specific if a condition is met.

lib/round.rb
```diff
@@ -0,0 +1,96 @@
...
+  def collect_category
+    deck.cards.each do |card|
+      @categories << card.category if @categories.include?(card.category) == false
+    end
+  end
+
+  def start
+    @total_cards = deck.cards.count
+    collect_category
```
megstang, line 67: This functionality doesn't feel like it belongs within the responsibility of this method. I wonder if the logic for collect_categories could make more sense in the Deck class instead of the Round class.

lib/round.rb
```diff
@@ -0,0 +1,96 @@
...
+    end_game
+  end
+
+  def end_game
+    puts "****** Game over! ******\n\n"
+    puts "You had #{correct_answers} correct guesses out of #{total_cards} \nfor a total score of %#{format('%.1f',
+                                                                                                            percent_correct)}.\n\n"
+    @categories.each do |category|
+      puts "#{category} - %#{format('%.1f', percent_correct_by_category(category))} correct"
+    end
```
megstang, line 93: All of this logic that's in here to run the game could possibly go in it's own class. For example, you could have a `Game` class that held all of these methods to run the game. That would help to adhere more to the OOP principle of SRP.

megstang (commented):
@ajkrumholz Nice work! I left some feedback for you to take a look over. Please let me know if you have any questions. i'll follow up in slack before the end of the day tomorrow with your scores.

PR #439, opened 2022-07-10
Chris Lyman

ChrisLyman:


lib/deck.rb
```diff
@@ -0,0 +1,14 @@
+class Deck
+
+  attr_reader :cards
+
+  def initialize(cards)
+    @cards = cards
+  end
+
```
epintozzi, line 8: The interaction pattern indicates we want to create a method called `count` for `Deck` specifically. Something like this should do the trick:
```ruby
def count
  @cards.count
end
```

It does feel a little wonky to have this since it's not doing a whole lot for us, but it can save us a step later on by including it.

lib/round.rb
```diff
@@ -0,0 +1,77 @@
...
+  def number_correct_by_category(category)
+    correct_by_category = 0
+    turns.each do |turn|
+      if turn.card.category == category
+        if turn.correct?
+          correct_by_category += 1
+        end
+      end
+    end
+    correct_by_category
```
epintozzi, line 38: Try to see if you can refactor this to use `count` instead of `each`

lib/round.rb
```diff
@@ -0,0 +1,77 @@
...
+    (@number_correct / turns.length.to_f) * 100.0
+  end
+
+  def percent_correct_by_category(category)
+    number_in_category = 0
+    turns.each do |turn|
+      if turn.card.category == category
+        number_in_category += 1
+      end
+    end
```
epintozzi, line 51: Other than incrementing `number_in_category` this portion of the method is very similar to `cards_in_category` method on `Deck`. You can reuse that method here since we assume we've gone through the whole deck of cards by the time we calculate percentage correct. That cleans up a little bit of redundancy in the code to get it DRYer.

lib/round.rb
```diff
@@ -0,0 +1,77 @@
...
+    puts "--------------------------------------------------------------------"
+    deck.cards.count.times do
+      show_card
+    end
+
+    puts"********************GAME OVER**********************"
+    puts "You had #{number_correct} guesses correct out of #{turns.count} for a total score of #{percent_correct}%"
+    puts "STEM - #{percent_correct_by_category(:STEM)}% Correct"
+    puts "Turing Staff - #{percent_correct_by_category(:"Turing Staff")}% Correct"
+    puts "Pop Culture - #{percent_correct_by_category(:PopCulture)}% Correct"
```
epintozzi, line 66: It's not ideal to hardcode the categories here. I would instead write a method that collects all the categories in a given deck and iterate through that list to create this output instead. That makes the code a lot more dynamic so we can add as many card categories as we want without having to change the code here in `Round` to account for it.

spec/card_generator_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    describe '#initialize' do
+      it 'should initialize an instance of CardGenerator' do
+        expect(@card_generator).to be_instance_of(CardGenerator)
+      end
+    end
+
+      it 'has an array of cards' do
+        expect(@card_generator.cards).to be_an Array
+      end 
+end
```
epintozzi, line 19: I'd probably also add a test that verifies the Array contains `Card` objects, just to make sure this class is actually doing what we want it to.

PR #440, opened 2022-07-10
Kevin Ta -Flashcards

KevinT001:


lib/turn.rb
```diff
@@ -0,0 +1,26 @@
...
+  def correct?
+    if @guess == card.answer
+      true
+    else
+      false
+    end
+  end
+
+  def feedback
+    if correct? == true
```
abdulredd, line 19: Great job using the `correct?` method here, but remember since the `correct?` method returns a boolean, we do not need to compare it to a boolean in an `if` statement. This method could have just been written like so:

```ruby
def feedback
  if correct?
    "Correct!"
  else
    "Incorrect."
  end
end
```

lib/deck.rb
```diff
@@ -0,0 +1,17 @@
...
+  def initialize(cards)
+    @cards = cards
+  end
+
+  def count
+    cards.length
+  end
+
+  def cards_in_category(category)
+    cards.find_all do |card|
```
abdulredd, line 13: Great job using `find_all` here instead of defaulting to `each`

lib/round.rb
```diff
@@ -0,0 +1,42 @@
+class Round
+  attr_reader :deck,
+              :turns
+  def initialize(deck)
+    @deck = deck
+    @turns = []
+  end
+
+  def current_card
+    @deck.cards[turns.count]
```
abdulredd, line 10: I like how you use the number of turns to keep track of what the current card should be, very smart.

lib/round.rb
```diff
@@ -0,0 +1,42 @@
...
+    ((number_correct.to_f / @turns.count) * 100).truncate()
+  end
+
+  def percent_correct_by_category(category)
+    (number_correct_by_category(category) / deck.cards_in_category(category).count.to_f) * 100
+  end
+
+  private
+
+  def correct_turns
```
abdulredd, line 38: Helper methods FTW. Great job breaking this method up into its own helper method. Even managing to hide it behind `private`.

lib/round.rb
```diff
@@ -0,0 +1,42 @@
...
+    new_turn
+  end
+
+  def number_correct
+    correct_turns.count
+
+  end
+
+  def number_correct_by_category(category)
+    correct_turns.select { |turn| turn.card.category == category }.count
```
abdulredd, line 25: You're demonstrating a command of Ruby mechanics by method chaining the `count` method onto your `select` block. Not many people know that you can do that.

lib/round.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def percent_correct
+    ((number_correct.to_f / @turns.count) * 100).truncate()
```
abdulredd, line 29: I would have much rather seen `round` here in place of `truncate` since with `truncate`, the numbers have the decimal just get chopped off, when `round` would round up or down accordingly.

spec/turn_spec.rb
```diff
@@ -0,0 +1,39 @@
+require './lib/card'
+require './lib/turn'
+
+
+RSpec.describe Turn do
+  card = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+  turn = Turn.new("Juneau", card)
+
+  it 'exists' do
+    expect(card).to be_a(Card)
```
abdulredd, line 10: We already made this check from inside the `card_spec.rb` file so doing it again here is just redundant.

spec/turn_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  it 'has a card' do
+    expect(turn.card).to eq(card)
+  end
+
+  it 'has a guess' do
+    expect(turn.guess).to eq("Juneau")
+  end
+
+  it 'can return a boolean for guess' do
```
abdulredd, line 22: When ever you have a method that can either return a `true` or `false` value, then you want to make sure that you are testing both sides of the condition so that you are sure that the method works in any scenario. The same thing goes for your `feedback` method.

spec/turn_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    expect(turn.correct?).to eq(true)
+  end
+
+  it 'can give feedback' do
+    expect(turn.feedback).to eq("Correct!")
+  end
+
+  it 'has a card with incorrect guess' do
```
abdulredd, line 30: I'm not sure I understand the point of this test since the expectation is the same as line 15.

PR #441, opened 2022-07-10
Riley McCullough

Rileybmcc:
Riley McCullough Flash Card Project 2207


flashcard_runner.rb
```diff
@@ -0,0 +1,62 @@
+require './lib/card'
+require './lib/turn'
+require './lib/deck'
+require './lib/round'
+
+@card_1 = Card.new("Who is the most recent dwarf to hold the title elf friend?", "Gimli", :LOTR)
```
abdulredd, line 6: Is there a reason why all these variable in your runner file are all instance variables and not just normal variables?

Rileybmcc: no particular reason, I had been writing tests and I absentmindedly added the @ symbols

PR #442, opened 2022-07-10
Kaelin Sleevi

KaelinSleevi:
Flash_cards project.

### DIFF
--- a/lib/deck.rb
+++ b/lib/deck.rb
@@ -1,0 +1,25 @@
+require './lib/card'
+require './lib/turn'
+class Deck
+  attr_reader :cards, :correct
+
+  def initialize(cards)
+    @cards = cards
+
+  end
+
+  def count
+    cards.count
+  end
+
+  def cards_in_category(category)
+    @cards.find_all do |card|
+      card.category == category
+    end
+  end
+
+  def correct?
+    @guess == card.answer
+  end
+
+end
--- a/lib/flash_runner.rb
+++ b/lib/flash_runner.rb
@@ -1,0 +1,44 @@
+require './lib/card'
+require './lib/turn'
+require './lib/deck'
+require './lib/round'
+
+
+@card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+@card_2 = Card.new("The Viking spacecraft sent back to Earth photographs and reports about the surface of which planet?", "Mars", :STEM)
+@card_3 = Card.new("Describe in words the exact direction that is 697.5° clockwise from due north?", "North north west", :STEM)
+@cards = [@card_1, @card_2, @card_3]
+@deck = Deck.new(@cards)
+@round = Round.new(@deck)
+
+def start
+
+  puts "Welcome! You're playing with #{@round.deck.cards.count} cards."
+  puts "--------------------------------------------------------------"
+  loop_cards
+end
+
+
+def loop_cards
+  number_loops = @round.deck.cards.count
+  number_loops.times do
+    card_loop = @round.turns.count + 1
+    puts "This is card number #{card_loop} out of #{@round.deck.count}."
+    puts "Question: #{@round.current_card.question}"
+    input = gets.chomp
+    @round.take_turn(input)
+    puts @round.turns.last.feedback
+  end
+
+  def end_of_game
+    puts "****** Game over! ******"
+    puts "You had #{@round.correct_turns.count} guesses out of #{@round.turns.count} score of #{@round.percent_correct}."
+    puts "#{(@round.percent_correct_by_category(:STEM))}% correct"
+    puts "#{(@round.percent_correct_by_category(:Geography))}% correct"
+  end
+end
+
+
+
+start
+end_of_game
--- a/lib/round.rb
+++ b/lib/round.rb
@@ -1,0 +1,54 @@
+require './lib/card'
+require './lib/turn'
+require './lib/deck'
+
+class Round
+  attr_reader :deck, :turns, :correct_turns
+
+  def initialize(deck)
+    @deck = deck
+    @turns = []
+    @correct_turns = []
+    #@round = @card_1
+  end
+
+  def current_card
+    #@round = Round.new(@deck)
+    @deck.cards[@turns.count]
+  end
+
+  def take_turn(guess)
+
+    turn_new = Turn.new(guess, current_card)
+     if turn_new.correct?
+       @correct_turns << turn_new
+     end
+    @turns << turn_new
+    turn_new
+  end
+
+  def number_correct
+    @correct_turns.count
+  end
+
+  def number_correct_by_category(category)
+    amount = @correct_turns.select do |turn|
+      turn.card.category == category
+    end
+    amount.count
+  end
+
+  def percent_correct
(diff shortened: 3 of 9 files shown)


lib/flash_runner.rb
```diff
@@ -0,0 +1,44 @@
...
+    @round.take_turn(input)
+    puts @round.turns.last.feedback
+  end
+
+  def end_of_game
+    puts "****** Game over! ******"
+    puts "You had #{@round.correct_turns.count} guesses out of #{@round.turns.count} score of #{@round.percent_correct}."
+    puts "#{(@round.percent_correct_by_category(:STEM))}% correct"
+    puts "#{(@round.percent_correct_by_category(:Geography))}% correct"
+  end
```
epintozzi, line 38: I thi

lib/flash_runner.rb
```diff
@@ -0,0 +1,44 @@
...
+  number_loops.times do
+    card_loop = @round.turns.count + 1
+    puts "This is card number #{card_loop} out of #{@round.deck.count}."
+    puts "Question: #{@round.current_card.question}"
+    input = gets.chomp
+    @round.take_turn(input)
+    puts @round.turns.last.feedback
+  end
+
+  def end_of_game
```
epintozzi, line 33: I think this is probably unintentional, but this `end_of_game` method is nested within the `loop_cards` method and we don't want to nest methods like that. The `end` on line 39 should be moved up to line 32 to fix this.

lib/flash_runner.rb
```diff
@@ -0,0 +1,44 @@
...
+    puts "****** Game over! ******"
+    puts "You had #{@round.correct_turns.count} guesses out of #{@round.turns.count} score of #{@round.percent_correct}."
+    puts "#{(@round.percent_correct_by_category(:STEM))}% correct"
+    puts "#{(@round.percent_correct_by_category(:Geography))}% correct"
```
epintozzi, line 37: Right now you are hardcoding the categories. It would be more dynamic if you wrote a method to collect a list of all categories in a given deck that you could then iterate through to calculate the percent correct.

lib/flash_runner.rb
```diff
@@ -0,0 +1,44 @@
...
+
+def start
+
+  puts "Welcome! You're playing with #{@round.deck.cards.count} cards."
+  puts "--------------------------------------------------------------"
+  loop_cards
+end
+
+
+def loop_cards
```
epintozzi, line 22: This is a good method! I like how dynamic it is and that it will change on its own depending on how many cards are in a deck.

spec/turn_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    turn = Turn.new("Juneau", card)
+
+    expect(turn.feedback).to eq("Correct!")
+  end
+
+  it 'incorrect? returns false if guess is incorrect' do
+    card = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    turn = Turn.new("Juneau", card)
+
+    expect(turn.correct?).to eq(true)
```
epintozzi, line 50: This is testing the same as line 36. Based on the title I think you might have been meaning to do something like `expect(turn.correct?).to eq(false)` which is a great idea. We want to test all possible outcomes for our code.

spec/turn_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    #expect(card).to be_instance_of(Card)
+  #end
+
+  #it 'can take a turn' do
+    #card = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    #turn = Turn.new("Juneau", card)
+
+    #expect(turn).to be_instance_of(Turn)
+  #end
+#end
```
epintozzi, line 85: Generally speaking it's best practice to cleanup any unused code before submitting a project/merging to main (or whatever the final destination may be).

lib/turn.rb
```diff
@@ -0,0 +1,36 @@
...
+  #end
+
+  #def guess
+    #@guess
+  #end
+
+  #def correct?
+    #@guess == @card.answer
+  #end
+
```
epintozzi, line 35: Clean up unused code before submitting

lib/deck.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def cards_in_category(category)
+    @cards.find_all do |card|
+      card.category == category
+    end
+  end
+
+  def correct?
+    @guess == card.answer
+  end
```
epintozzi, line 23: I don't think this method is used for anything so we can get rid of it.

lib/deck.rb
```diff
@@ -0,0 +1,25 @@
...
+    @cards = cards
+
+  end
+
+  def count
+    cards.count
+  end
+
+  def cards_in_category(category)
+    @cards.find_all do |card|
```
epintozzi, line 16: Great use of the `find_all` enumerable method!

lib/round.rb
```diff
@@ -0,0 +1,54 @@
...
+  def percent_correct
+    num_1 = @correct_turns.count.to_f
+    num_2 = @turns.count.to_f
+
+    (num_1 / num_2) * 100
+  end
+
+  def percent_correct_by_category(category)
+    num_1 = @correct_turns.select {|turn| turn.card.category == category}.count.to_f
+    num_2 = @turns.select {|turn| turn.card.category == category}.count.to_f
```
epintozzi, line 50: The code for `num_2` is very similar to our `cards_in_category` method on `Deck`. You can use that here to reduce redundancy.

lib/round.rb
```diff
@@ -0,0 +1,54 @@
...
+require './lib/deck'
+
+class Round
+  attr_reader :deck, :turns, :correct_turns
+
+  def initialize(deck)
+    @deck = deck
+    @turns = []
+    @correct_turns = []
+    #@round = @card_1
```
epintozzi, line 12: Clean out unused code. (Also on line 16).

spec/round_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+  @round.take_turn("Venus")
+
+  expect(@round.percent_correct).to eq(50.0)
+end
+
+it 'tells you the percent correct by category' do
+  @new_turn = @round.take_turn("Juneau")
+  @round.take_turn("Venus")
+
+  expect(@round.percent_correct_by_category(:Geography)).to eq(100.0)
```
epintozzi, line 95: You have this setup to make two assertions, and I would follow through with that by adding the following to verify you aren't getting a false positive for one category.
`expect(@round.percent_correct_by_category(:STEM)).to eq(0.0)`

spec/round_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+
+  it 'tells you what turn you are on' do
+    @new_turn = @round.take_turn("Juneau")
+
+    expect(@round.turns).to eq([@new_turn])
+  end
+
+it 'tell you how many you guess correct' do
+   @new_turn = @round.take_turn("Juneau")
+
```
epintozzi, line 59: I would take another turn with an incorrect guess to verify the `number_correct` is in fact only incrementing on correct guesses. This could be a false positive (I don't think it is, but it COULD be if we aren't careful or the code changes).

PR #443, opened 2022-07-10
Thomas Hayes

thayes87:


flashcard_runner.rb
```diff
@@ -0,0 +1,48 @@
...
+puts card_3.question
+user_guess_3 = gets.chomp
+turn_3 = round.take_turn(user_guess_3)
+puts turn_3.feedback
+#Round 4
+puts "This is the last card"
+puts card_4.question
+user_guess_4 = gets.chomp 
+turn_4 = round.take_turn(user_guess_4)
+puts turn_4.feedback
```
epintozzi, line 39: Notice how the code for each round is identical except for the opening phrase ("This is card 2 out of 4") and the number in the variable name (card_1, card_2, etc). This would be a super fun refactor to see if you can figure out how to loop through the cards so that this is dynamic enough to handle any number of cards without having to update the runner to match.

flashcard_runner.rb
```diff
@@ -0,0 +1,48 @@
...
+
+puts "****** Game over! *******" 
+results = round.number_correct
+results_pct = round.percent_correct
+puts "You had #{results} correct guesses out of 4 for a total score of #{results_pct}%."
+pcbc_history = round.percent_correct_by_category(:History)
+pcbc_culture = round.percent_correct_by_category(:Culture)
+puts "History: #{pcbc_history}% correct"
+puts "Culture: #{pcbc_culture}% correct"
```
epintozzi, line 48: Similar to the above, think about how we might be able to write a method to give us a collection of all categories in a given deck so that this can be more dynamic rather than hardcoding 'history' and 'culture' in these four lines.

lib/turn.rb
```diff
@@ -0,0 +1,23 @@
+class Turn 
+  attr_reader :guess, :card 
+  def initialize(guess, card)
+    @guess = guess
+    @card = card
+  end
+
+  def correct?
+    if guess.downcase == card.answer.downcase
```
epintozzi, line 9: 🔥 Excellent that you've considered a user might not have use proper caps

spec/turn_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    expect(turn_2.feedback). to eq("Incorrect!")
+  end
+
+  it 'is correct regardless of capitalization' do
+    card_2 = Card.new("Which planet is closest to the sun?", "Mercury", :STEM)
+    turn_2 = Turn.new("mercury", card_2)
+
+    expect(turn_2.correct?).to be true
+  end
```
epintozzi, line 53: And you tested for caps?! Great work!

spec/turn_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  it 'provides Correct! feedback' do
+    card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    turn_1 = Turn.new("Juneau", card_1)
+
+    expect(turn_1.feedback). to eq("Correct!")
+  end
+
+  it 'provides Incorrect feedback' do
```
epintozzi, line 41: This is great. It's good to always test every outcome, not just the "positive" or correct outcome. I might also add a line to explicitly test the `correct?` method as well (even though it is implicitly tested here). Generally every method we write should have its own test.

lib/deck.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  def count
+    count = cards.count
+  end
+  
+  def cards_in_category(name)
+    groups = cards.group_by do |card| 
+      card.category 
+    end
+    groups[name]
```
epintozzi, line 15: Nice use of `group_by`. I like this implementation! Other enumerable methods that work well here are `select` and `find_all`

lib/round.rb
```diff
@@ -0,0 +1,53 @@
+class Round
+  attr_reader :deck, :turns, :correct_answers
+  def initialize(deck)
+     @deck = deck
+     @turns = []
+     @correct_answers = 0
+  end
+
+  def current_card 
+    deck.cards.shift
```
epintozzi, line 10: I might consider doing the actual shift in the `take_turn` method so that the deck is only changing after a turn is done. Right now, anytime the `current_card` method is called, it will shift the deck. The way the program is set up, this seems to work fine, but if you ever called current_card more than once between turns, I think you'd end up skipping cards. That way current card method is only responsible for returning the top card on the deck and nothing more.

PR #444, opened 2022-07-10
Rebecka Hendricks

rebeckahendricks:


flashcard_runner.rb
```diff
@@ -0,0 +1,19 @@
...
+card_5 = Card.new("What nationality was Picasso?", "Spanish", :Art)
+card_6 = Card.new("What is stored in a camel's hump?", "fat", :Animals)
+card_7 = Card.new("What color is a giraffe's tongue?", "blue", :Animals)
+card_8 = Card.new("Which is the world's largest lizard?", "Komodo dragon", :Animals)
+
+cards = [card_1, card_2, card_3, card_4, card_5, card_6, card_7, card_8]
+deck = Deck.new(cards)
+round = Round.new(deck)
+
+round.start
```
megstang, line 19: Runner file looks good!

flashcard_runner_it4.rb
```diff
@@ -0,0 +1,13 @@
...
+require './lib/turn'
+require './lib/deck'
+require './lib/round'
+require './lib/card_generator'
+
+cards = CardGenerator.new('.lib/cards.txt').cards
+deck = Deck.new(cards)
+round = Round.new(deck)
+
+round.start
```
megstang, line 11: This, as well as the CardGenerator class's pseudocode is a great start to Iteration 4. I liked seeing you write out comments for what you knew you'd need to do in that method.

lib/card.rb
```diff
@@ -0,0 +1,12 @@
...
+              :answer,
+              :category
+
+  def initialize(question, answer, category)
+    @question = question
+    @answer = answer
+    @category = category
+  end
+
+end
```

lib/deck.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def cards_in_category(category)
+    cards_category = []
+    @cards.each do |card|
+      if card.category == category
+        cards_category << card
+      end
+    end
+    cards_category
+  end
```
megstang, line 22: Today, we'll talk more about other enums you can use. You did a great job using each here, and this would be a good example of a time to use find_all.

lib/round.rb
```diff
@@ -0,0 +1,93 @@
...
+
+    if new_turn.correct? == true
+      @number_correct += 1
+    end
+
+    def feedback
+      if new_turn.correct? == false
+        "Incorrect."
+      end
+    end
```
megstang, line 31: This method is nested in your take_turn method, and doesn't seem to ever be used. The only time you ever call the method `feedback` is on a Turn object.

lib/round.rb
```diff
@@ -0,0 +1,93 @@
...
+    @deck.cards.shift#Change current card to next card in array - push out first element
+
+    new_turn
+  end
+
+  def number_correct_by_category(category_turn)
+    category_correct = 0
+
+    @turns.each do |turn|
+      if turn.card.category == category_turn && turn.correct? == true
```
megstang, line 41: instead of `turn.correct? == true` you can just leave it as `turn.correct?` since that will automatically return true or false.

lib/turn.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def feedback
+    if @guess == card.answer
+      "Correct!"
+    else
+      "Incorrect."
+    end
+  end
+
+end
```
megstang, line 24: 👍

lib/turn.rb
```diff
@@ -0,0 +1,24 @@
...
+  attr_reader :guess,
+              :card
+
+  def initialize(guess, card)
+    @guess = guess
+    @card = card
+  end
+
+  def correct?
+    @guess == card.answer
```
megstang, line 13: reference `@card` instead of `card` the attr_Reader.

megstang (commented):
@rebeckahendricks Nice work on this! Please let me know if you have any questions on my feedback. I'll be following up on slack later today with your scores.

PR #445, opened 2022-07-10
Erik Riggs

eriggs0207:


lib/turn.rb
```diff
@@ -0,0 +1,23 @@
...
+    @card = card
+  end
+
+  def correct?
+    @guess == @card.answer
+    
+  end
+
+  def feedback
+    if self.correct? == true
```
abdulredd, line 17: You do not need to use `self` here and it is already implied. Also since we know that a call to `correct?` will either return `true` or `false`, we do not need to explicitly compare this to a `true` or `false` value and can get away with doing something like:

```ruby
def feedback
  if correct?
    return "Correct!"
  else
    return "Incorrect."
  end
end
```

lib/deck.rb
```diff
@@ -0,0 +1,17 @@
+class Deck
```
abdulredd, line 1: According to the specification, the `Deck` class is supposed to also have a `count` method, this method seems to be missing.

lib/round.rb
```diff
@@ -0,0 +1,61 @@
...
+              :number_correct
+
+  def initialize(deck)
+    @deck = deck
+    @turns = []
+    @number_correct = []
+  end
+
+  def current_card
+    deck.cards[turns.count]
```
abdulredd, line 16: I like that use use the number of turns to determine what the current card should be.

lib/round.rb
```diff
@@ -0,0 +1,61 @@
...
+    new_turn
+  end
+
+  def number_correct
+    @number_correct.count
+  end
+
+  def number_correct_by_category(category)
+    correct_by_category = turns.select do |turn|
+      turn.card.answer == turn.guess && turn.card.category == category
```
abdulredd, line 35: This portion of code looks familiar `turn.card.answer == turn.guess`, don't we have something like this already defined inside the `correct?` method in the `Turn` class? How can we use that method to help us not repeat ourselves here?

```ruby
correct_by_category = turns.select do |turn|
  turn.correct? && turn.card.category == category
end
```

lib/round.rb
```diff
@@ -0,0 +1,61 @@
...
+    correct_by_category.count
+
+  end
+
+  def percent_correct
+   percent = (@number_correct.count.to_f / @turns.count.to_f) * 100
+   percent.round
+  end
+
+  def percent_correct_by_category(category)
```
abdulredd, line 46: The indentation in this method is a little off. At first glance it looks like the `end` on line 50 is ending the method with it is actually the `end` that belongs to the `select` block.

spec/deck_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  it 'should count cards in deck' do
+    card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    card_2 = Card.new("The Viking spacecraft sent back to Earth photographs and reports about the surface of which planet?", "Mars", :STEM)
+    card_3 = Card.new("Describe in words the exact direction that is 697.5° clockwise from due north?", "North north west", :STEM)
+
+    cards = [card_1, card_2, card_3]
+    deck = Deck.new(cards)
+
+    expect(cards.count).to eq(3)
```
abdulredd, line 35: The expectation you want to check here, according to the specification is `deck.count`, having done this you would have realized that you are missing a `count` method for the `Deck` class.

spec/deck_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+  
+  it 'should return cards in category' do
```
abdulredd, line 38: For this last test you want to also make sure that if a category that you do not have is entered into the method that an empty array is returned.

`expect(deck.cards_in_category(:Finance)).to eq([])`

spec/round_spec.rb
```diff
@@ -0,0 +1,154 @@
...
+    card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    card_2 = Card.new("The Viking spacecraft sent back to Earth photographs and reports about the surface of which planet?", "Mars", :STEM)
+    card_3 = Card.new("Describe in words the exact direction that is 697.5° clockwise from due north?", "North north west", :STEM)
+
+    cards = [card_1, card_2, card_3]
+    deck = Deck.new(cards)
+    round = Round.new(deck)
+
+    new_turn = round.take_turn("Juneau")
+    expect(round.turns).to include(new_turn)
```
abdulredd, line 72: Great use of the `include` matcher here.

PR #446, opened 2022-07-10
Stephen Fabian

stephenfabian:


flashcard_runner.rb
```diff
@@ -0,0 +1,18 @@
...
+card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+card_2 = Card.new("The Viking spacecraft sent back to Earth photographs and reports about the surface of which planet?", "Mars", :STEM)
+card_3 = Card.new("Describe in words the exact direction that is 697.5° clockwise from due north?", "North north west", :STEM)
+deck = Deck.new([card_1, card_2, card_3])
+round = Round.new(deck)
+
+
+round.start
+
+
```
megstang, line 18: Runner file looks good!

lib/card.rb
```diff
@@ -0,0 +1,13 @@
+class Card
+    
+attr_reader :question, :answer, :category
+
+   def initialize(question, answer, category)
+    @question = question
+    @answer = answer
+    @category = category
+    end 
+end
```

lib/deck.rb
```diff
@@ -0,0 +1,20 @@
...
+    cards.count
+end
+
+def cards_in_category(category)
+@cards.find_all do |card|
+card.category == category
+end
+end
+
+end
```
megstang, line 20: Careful with your indentation. make sure def's, and tabbed over, and also the content inside of them: 
```ruby
class Deck 

       def initialize()
               code code code 
       end 

end
```

lib/round.rb
```diff
@@ -0,0 +1,111 @@
...
+
+def current_card
+    @deck.cards[@turns.count]
+end
+
+def take_turn(guess)
+    new_turn = Turn.new(guess, current_card)
+    @turns << new_turn
+    if new_turn.correct?
+        @num_correct << new_turn
```
megstang, line 20: Instead of putting the turns into this num_correct, we could increment the value of num_correct. For example, we could set `@num_correct = 0` in the initialize method, and then on this line, we can do `@num_correct += 1`

megstang: If you did that ^^, then you could rename `@num_correct` to be `@number_correct`, replace your attr_reader for `:number_correct`, and then you won't need to make your own method for number_correct.

lib/round.rb
```diff
@@ -0,0 +1,111 @@
...
+
+def number_correct_by_category(category)
+    num_correct_in_category = []
+@turns.each do |turn|
+    if turn.correct? && turn.card.category == category
+        num_correct_in_category << turn
+    end
+end
+
+num_correct_in_category.count
```
megstang, line 38: Similarly here, you could have your accumulator on line 31 be set to 0 instead of [], then you don't ned to count up the elements on line 38.

megstang: It just seems like you're storing more than you need to in these variables. Objects in an array take up more space than a single integer. And if you don't need those objects, and rather the count, you might as well start with an integer.

lib/round.rb
```diff
@@ -0,0 +1,111 @@
...
+
+def total_number_of_cards_in_category(category)
+    number_of_cards_in_category = []
+    turns.select do |turn|
+       if turn.card.category == category
+        number_of_cards_in_category << turn
+        end
+    end
+    number_of_cards_in_category.count
+end
```
megstang, line 52: You're using `select` AND a placeholder, which kind of makes the use of `select` unnecessary. We use `select` to return a subset of the original array. To still use `select`, you could do something like: 
```ruby 
    cards_in_category = turns.select do |turn|
               turn.card.category == category
     end.count
```

lib/round.rb
```diff
@@ -0,0 +1,111 @@
...
+turns.each do |turn|
+    card_categories << turn.card.category
+end
+
+
+(card_categories.uniq).each do |category|
+   p "#{category} - #{(number_correct_by_category(category) / total_number_of_cards_in_category(category)) * 100} % correct"
+#    p "#{category} - #{number_correct_by_category(category)} / #{total_number_of_cards_in_category}"
+
+end
```
megstang, line 92: This start method looks good as far as my expectation on this project (at this point in the module) goes. If we wanted to break this down a bit more, I think we could put this logic into another class. Maybe a `Game` class. And that `Game` class can be responsible for the flow/run of the game. In the game class, you could have a few methods, maybe one to welcome the player, one to loop through the game, and one to calculate the final outcome.

lib/turn.rb
```diff
@@ -0,0 +1,23 @@
...
+def feedback
+if correct?
+   "Correct!"
+else 
+    "Incorrect."
+end
+end
+
+end
+
```
megstang, line 23: 👍

megstang (commented):
@stephenfabian Nice work! Please let me know if you have any questions on my comments. I'll follow up in slack later today with your scores.

PR #448, opened 2022-07-10
Aleisha Mork

aleish-m:


flashcard_runner.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  puts " "
+  card_categories = []
+  round.turns.each do |turn|
+    card_categories << turn.card.category
+  end
+  (card_categories.uniq).each do |category|
+    puts "#{category} - #{round.percent_correct_by_category(category).to_i}% correct"
+  end
+end
```
megstang, line 53: You had asked for feedback around your runner file and the conventions around it. At this point in your developer career, I'm happy with what you were able to produce, and how you were able to get your game to run via this runner file. In the future, we'll challenge you to store as little code as possible outside of a class. Right now, you have code in this file that is not in a class. So, if you were to follow that convention, you could have a line of code in your runner file that was something like `Game.start`. Where `Game` is a new class that you created for the purpose of running the game. And `start` was a method that kicked off the flow of the game. The idea is that the flow of the game is stored in methods in that game class.

lib/card.rb
```diff
@@ -0,0 +1,11 @@
...
+
+attr_reader :question, :answer, :category
+def initialize (question, answer, category)
+  @question = question
+  @answer = answer
+  @category = category
+end
+
+
+end
```

lib/deck.rb
```diff
@@ -0,0 +1,24 @@
...
+    card_by_category = []
+    @cards.each do |card|
+      if card.category == (category)
+        card_by_category << card
+      end
+    end
+    card_by_category
+  end
+
+end
```
megstang, line 24: 👍 Nice job here!

lib/round.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def take_turn(guess)
+    new_turn = Turn.new(guess, current_card)
+      @turns << new_turn
+      if new_turn.correct?
+        @number_correct += 1
+      end
+    @deck.cards.shift
+    new_turn
+  end
```
megstang, line 25: Nice job here, this was a tough method!

lib/round.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def number_correct_by_category(card_category)
+    category_count = 0.0
+    @turns.each do |turn|
+      if turn.card.category == card_category && turn.correct?
```
megstang, line 30: nice use of the `correct?` method here.

lib/round.rb
```diff
@@ -0,0 +1,52 @@
...
+        category_count += 1
+      end
+    end
+    category_count
+  end
+
+  def percent_correct
+    (number_correct / turns.count) * 100
```
megstang, line 38: If you can call to the instance variable instead of the attr_Reader, try to remember to do that. When you call to `number_correct`, you're calling to the attr_reader, which returns the instance variable. But you can call directly to the instance variable and cut out the middle man instead.

lib/round.rb
```diff
@@ -0,0 +1,52 @@
...
+  def percent_correct_by_category(card_category)
+    category_total = 0.0
+    @turns.each do |turn|
+      if turn.card.category == card_category
+        category_total += 1
+      end
+    end
+    category_total
+    (number_correct_by_category(card_category)/category_total) * 100
+  end
```
megstang, line 50: 👍 Great work!

megstang (commented):
@aleish-m Really great work! Please let me know if you have any questions on any of the comments I left. I'm going to follow up by the end of the day tomorrow (via slack) with your scores.

PR #449, opened 2022-07-10
Benjamin Randolph

neb417:


lib/turn.rb
```diff
@@ -0,0 +1,17 @@
...
+    def initialize(guess, card)
+        @card = card
+        @guess = guess
+    end
+
+    def correct?
+      @guess == @card.answer
+    end
+
+    def feedback
```
abdulredd, line 13: I love the whole setup of this method and how you're using implicit and explicit returns to your advantage. The only thing I would add is that because `correct?` returns a boolean you do not actually need to compare it to a boolean and you can get away with doing something like this:

```ruby
def feedback
  return "Correct!" if correct?
  "Incorrect."
end
```

lib/deck.rb
```diff
@@ -0,0 +1,19 @@
+class Deck
```
abdulredd, line 1: This class is missing a `count` method and required from the specification. I believe this is what `cards_in_deck ` is supposed to be, but the name doesn't match what is in the specification.

lib/deck.rb
```diff
@@ -0,0 +1,19 @@
...
+    attr_reader :cards, :category
```
abdulredd, line 2: You have an `attr_reader` for `category` but I do not see an instance variable called `category` anywhere in the class.

lib/round.rb
```diff
@@ -0,0 +1,52 @@
+class Round
+    attr_reader :deck, :turns, :cards, :number_correct
+    
+    def initialize (deck)
+        @deck = deck
+        @cards = cards
```
abdulredd, line 6: I do not understand this line, `@cards = cards` what is `cards` and where is it coming from? if you put a `pry` in here, `@cards` is `nil` and stays `nil` as there is no code within this class or in any other file that makes it anything else. You also have an `attr_reader` for it but it is never used outside of this class. My guess is that you could probably get rid of it without any issues.

lib/round.rb
```diff
@@ -0,0 +1,52 @@
...
+        @turns = []
+        @number_correct = 0
+        @correct_turns = @correct_turns
```
abdulredd, line 9: We have a case of inception going on in lines 9 and 10 where we are setting the instance variable to be .. itself? I'm making an assumption, but I think what you are probably trying to do it start these values off as `nil` maybe and define them later on? Lines 30, 42, 46?

Well you can always start these instance variables off as nil, `@instance_variable = nil` or you can just leave them out all together and just define them when you need them in the methods down below where they get defined. While defining instance variables in your initialize is great because you can see them all in one location, this is not the only location where you can define them. They can be defined in any method of the class, the only caveat being you won't have that instance variable available until after the method that defines it gets called.

lib/round.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+
+    def current_card
+        deck.cards[0]
+    end
+
+    def take_turn(guess)
+        turn = Turn.new(guess, current_card)
+        @turns << turn
+        if guess == current_card.answer
```
abdulredd, line 21: How about instead of `if guess == current_card.answer` we do `if turn.correct?`? Would that still work?

lib/round.rb
```diff
@@ -0,0 +1,52 @@
...
+            @number_correct += 1 
+        end
+        deck.cards.rotate!
+        return turn
+    end
+
+    def number_correct_by_category(category)
+
+        @correct_turns = turns.select do |turn|
```
abdulredd, line 30: This method works just fine as written but just as an alternate to this method, you could also just chain on the `count` method directly to the block like so:

```ruby
def number_correct_by_category(category)
  turns.select do |turn|
    turn.card.category == category && turn.correct?
  end.count # <====
end
```

We can do this because the `select` enumerable returns and array of selected items, and then we are just calling `count` on that returned array.

flashcards_runner.rb
```diff
@@ -0,0 +1,60 @@
...
+puts ""
+puts "You had #{round.number_correct} correct guesses out of #{cards.count} for a total score of #{round.percent_correct}%"
+
+
+turns.each do |turn|
+    cat_arry << turn.card.category unless cat_arry.include?(turn.card.category)
+end
+
+cat_arry.each do |cat|
+    puts "#{cat} - #{round.percent_correct_by_category(cat)}%"
```
abdulredd, line 56: I like the fact that you made your solution dynamic and it can take any cards of any category and still produce proper output.

lib/deck.rb
```diff
@@ -0,0 +1,19 @@
...
+        @cards = cards
+    end
+
+    def cards_in_deck
+        cards.count
+    end
+
+    def cards_in_category(category)
+        card_count = []
+        cards.each do |card|
```
abdulredd, line 14: Is there another method other than `each` that could have probably been more suitable here?

spec/turns_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+        expect(@turn.card).to eq(@card)
+    end
+
+    it 'returns a guess' do
+
+        expect(@turn.guess).to eq("Saturn")
+    end
+
+    it 'is_guess_correct?' do
+
```
abdulredd, line 27: i think it would benefit you to not only check for when a guess is incorrect but also when a guess if correct as well to make sure the methods `correct?` and `feedback` are both doing what is expected in both scenarios.

spec/round_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+
+        expect(new_turn.class).to eq(Turn)
+        expect(new_turn.correct?).to eq(true)
+        expect(@round.turns).to eq(@turns)
+        expect(@round.number_correct).to eq(1)
+    end
+
+    it '6. moves to the next card' do
+        new_turn = @round.take_turn("Juneau")
+        @turns = [new_turn]
```
abdulredd, line 53: Line 53 and 65 don't seem to be needed for either test.

I do love the robustness of each test, however, and how each has several expectations!

spec/round_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+        expect(@round.current_card).to eq(@card_1)
+    end
+
+    it '5. takes a turn' do
+        new_turn = @round.take_turn("Juneau")
+        @turns = [new_turn]
+
+        expect(new_turn.class).to eq(Turn)
+        expect(new_turn.correct?).to eq(true)
+        expect(@round.turns).to eq(@turns)
```
abdulredd, line 47: I don't think you need the `@turns` variable on line 43. This expectation could have just been written like `expect(@round.turns).to eq([new_turn])` and the result would have been the same. And since this is the only place where you seem to be using `@turns` you could then just remove it in the `before :each` block and every where else.

PR #450, opened 2022-07-10
Alex Mora

AlexMR-93:


flashcard_runner.rb
```diff
@@ -0,0 +1,35 @@
...
+    round.take_turn(guess)
+    round.turns.last.correct?
+    puts(round.turns.last.feedback)
+  end
+
+  puts(("Game Over!".center(40, "*")))
+  puts(("you had #{round.number_correct} out of #{deck.count} for a total #{round.percent_correct.truncate}%"))
+  puts("STEM - #{round.percent_correct_by_category(:STEM).truncate}% correct")
+  puts("PopCulture - #{round.percent_correct_by_category(:PopCulture).truncate}% correct")
+  puts("TuringStaff - #{round.percent_correct_by_category(:TuringStaff).truncate}% correct")
```
epintozzi, line 32: Ideally we want our categories to be dynamic and not hardcoded so that the cards in our deck can change without having to update the code here in the runner. I would probably start by writing a method that returns a collection of categories in a given deck and iterate over those categories to return the scores for each one.

spec/turn_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  it("has feedback") do
+    expect(@turn.feedback).to(eq("Correct!"))
+  end
+
+  it("has card w. answer equal to mercury") do
+    @card = Card.new("Which planet is closest to the sun?", "Mercury", :STEM)
+    @turn = Turn.new("Saturn", @card)
+    expect(@turn.card).to(eq(@card))
+  end
```
epintozzi, line 35: We should also test the `correct?` and `feedback` methods for a wrong answer. You can add that to this test here with something like:
```ruby
    expect(@turn.correct?).to(eq(false))
    expect(@turn.feedback).to(eq("Incorrect."))
```

lib/deck.rb
```diff
@@ -0,0 +1,17 @@
...
+  def initialize(cards)
+    @cards = cards
+  end
+
+  def count
+    @cards.length
+  end
+
+  def cards_in_category(category)
+    cards.select do |card|
```
epintozzi, line 13: Great use of the `select` method 🔥

lib/round.rb
```diff
@@ -0,0 +1,42 @@
+class Round
+  attr_reader :deck, :turns, :number_correct
```
epintozzi, line 2: We don't need the `number_correct` here because you have defined this method below on line 27.

lib/round.rb
```diff
@@ -0,0 +1,42 @@
...
+  def number_correct_by_category(category)
+    @right_answers.select { |card| card.category == category }.count
+  end
+
+  def percent_correct
+    (@number_correct.to_f) / (@turns.count.to_f) * 100
+  end
+
+  def percent_correct_by_category(category)
+    (number_correct_by_category(category).to_f / deck.cards_in_category(category).count) * 100
```
epintozzi, line 40: Really glad to see you've re-used the `cards_in_category` method that you wrote in your `Deck` class. Great work!

lib/round.rb
```diff
@@ -0,0 +1,42 @@
...
+      @number_correct += 1
+      @right_answers << @curr_card
+    end
+
+    @new_turn
+  end
+
+  def number_correct
+    @right_answers.count
+  end
```
epintozzi, line 29: I feel like `number_correct`, `@number_correct`, and `@right_answers` are all doing a little bit of the same thing with some very minor variances. I'd maybe try to refactor to see if you can consolidate that a little bit.

PR #451, opened 2022-07-10
Rob Schlereth

rschlereth:


rschlereth:
Had some trouble with first bunch of commits not showing up. As best I can tell, it was an issue with my having cloned this and made commits during last (Mod1) inning.

flashcard_runner.rb
```diff
@@ -0,0 +1,33 @@
+require './lib/card'
+require './lib/turn'
+require './lib/deck'
+require './lib/round'
+
+def start(round)
+    puts "Welcome! You're playing with 4 cards."
```
epintozzi, line 7: You're using `round.deck.count` on line 11, but could also use it here so you aren't hardcoding the number of cards in the game.

flashcard_runner.rb
```diff
@@ -0,0 +1,33 @@
...
+    puts "Question: #{round.current_card.question}"
+    guess = gets.chomp
+    answer = round.take_turn(guess)
+    puts answer.feedback
+    end
+    puts "****** Game over! ******"
+    puts "You had #{round.number_correct} out of #{round.deck.count} for a total score of #{round.percent_correct.round(1)}."
+    puts "Geography - #{round.percent_correct_by_category(:Geography)}"
+    puts "STEM - #{round.percent_correct_by_category(:STEM)}"
+end
```
epintozzi, line 21: For a refactor, I would try to create a method that collects a list of all our categories so we can iterate through that dynamically to present the scores, rather than hardcoding it.

lib/turn.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  attr_reader :guess, :card
+  def initialize(guess, card)
+    @guess = guess
+    @card = card
+
+  end
+
+  def correct?
+    @guess == @card.answer
```
epintozzi, line 11: Great! I really love that your solution is so short and elegant (rather than building this out with a longer `if` statement).

spec/turn_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  it "exists" do
+    card = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    turn = Turn.new("Juneau", card)
+    expect(turn).to be_a(Turn)
+  end
+
+  it "has readable attributes" do
+    card = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    turn = Turn.new('Juneau', card)
+  end
```
epintozzi, line 15: This isn't actually testing anything because there's no `expect` statement. I'm guessing you meant to put something like this here:
`expect(turn.card).to eq(card)`

lib/deck.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def cards_in_category(category)
+    matches = []
+    @cards.each do |card|
+      if card.category == category
+        matches << card
+      end
+    end
+    matches
```
epintozzi, line 19: This works! We can also use `select` or `find_all` here to make it even more concise:
```ruby
    @cards.select do |card|
      card.category == category
    end
```

spec/round_spec.rb
```diff
@@ -0,0 +1,182 @@
...
+        card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+        card_2 = Card.new("The Viking spacecraft sent back to Earth photographs and reports about the surface of which planet?", "Mars", :STEM)
+        card_3 = Card.new("Describe in words the exact direction that is 697.5° clockwise from due north?", "North north west", :STEM)
+        deck = Deck.new([card_1, card_2, card_3])
+        round = Round.new(deck)
+        new_turn = round.take_turn("Juneau")
+        next_turn = round.take_turn("Venus")
+        following_turn = round.take_turn("North north west")
+        expect(round.percent_correct_by_category(:Geography)).to eq(100.0)
+
```
epintozzi, line 169: If we add another assertion here to make sure it's working correctly for percentages less than 100, we will see that this method isn't working quite as intended.
`expect(round.percent_correct_by_category(:STEM)).to eq(50.0)`

lib/round.rb
```diff
@@ -0,0 +1,51 @@
+class Round
+    attr_reader :deck, :turns, :current_card, :card_index
```
epintozzi, line 2: We don't need a `current_card` `attr_reader` because you have defined that method below.

lib/round.rb
```diff
@@ -0,0 +1,51 @@
...
+    end
+
+    def number_correct
+        number_correct = 0
+        @turns.each do |turn|
+            if turn.correct?
+                then number_correct += 1
+            end
+        end
+    number_correct
```
epintozzi, line 28: Try to see if you can refactor this using the `count` enumerable method.

lib/round.rb
```diff
@@ -0,0 +1,51 @@
...
+        correct_by_category
+        
+    end
+
+    def percent_correct
+        number_correct.to_f / turns.length * 100
+    end
+
+    def percent_correct_by_category(category)
+        number_correct_by_category(category) / @deck.cards_in_category(category).length * 100
```
epintozzi, line 48: Glad to see you using the `cards_in_category` method here rather than re-writing similar code. Nice job!

epintozzi, line 48: Also if you add a `.to_f` here, it should solve the issue I noted in the spec about it not working quite as intended. You might have to round again after you get your result to match the functionality exactly, but you'll be a lot closer!

PR #452, opened 2022-07-10
Kyongnam Kang

kangkna:


lib/deck.rb
```diff
@@ -0,0 +1,19 @@
...
+      cat_arr = []
+
+      cards.each do |card|
+        if card.category == category
+        cat_arr << card
+      end
+    end
+
+    cat_arr
+    end
```
megstang, line 18: Keep your eye on syntax. We always want to make sure our end's match up with their corresponding block starter (def, if, each, etc.)

spec/deck_spec.rb
```diff
@@ -0,0 +1,41 @@
+require './lib/card'
+require './lib/deck'
+
+RSpec.describe Card do
```
megstang, line 4: This test should describe the `Deck` class, not `Card`

spec/deck_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  it 'has cards' do
+    card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    card_2 = Card.new("The Viking spacecraft sent back to Earth photographs and reports about the surface of which planet?", "Mars", :STEM)
+    card_3 = Card.new("Describe in words the exact direction that is 697.5° clockwise from due north?", "North north west", :STEM)
+
+    cards = [card_1, card_2, card_3]
+  end
```
megstang, line 11: There is no assertion in this method

spec/deck_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  it 'can catergorize decks' do
+    card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    card_2 = Card.new("The Viking spacecraft sent back to Earth photographs and reports about the surface of which planet?", "Mars", :STEM)
+    card_3 = Card.new("Describe in words the exact direction that is 697.5° clockwise from due north?", "North north west", :STEM)
+
+    deck = Deck.new(cards)
+    expect(deck.cards_in_category(:STEM)).to eq([card_2, card_3])
+    expect(deck.cards_in_category(:Geography)).to eq([card_1])
+    expect(deck.cards_in_category("Pop Culture")).to eq([])
+  end
```
megstang, line 40: these tests look good 👍

spec/round_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+RSpec.describe Round do
+  it 'exists' do
+    card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    card_2 = Card.new("The Viking spacecraft sent back to Earth photographs and reports about the surface of which planet?", "Mars", :STEM)
+    card_3 = Card.new("Describe in words the exact direction that is 697.5° clockwise from due north?", "North north west", :STEM)
+
+    deck = Deck.new([card_1, card_2, card_3])
+    round = Round.new(deck)
+
+    expect(deck).to be_instance_of(Deck)
```
megstang, line 15: no need for this assertion in this spec since we're testing the Round class in this spec.

spec/round_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  it "has a deck" do
+    card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    card_2 = Card.new("The Viking spacecraft sent back to Earth photographs and reports about the surface of which planet?", "Mars", :STEM)
+    card_3 = Card.new("Describe in words the exact direction that is 697.5° clockwise from due north?", "North north west", :STEM)
+
+    deck = Deck.new([card_1, card_2, card_3])
+    round = Round.new(deck)
+
+    expect(round.deck).to eq([card_1, card_2, card_3])
```
megstang, line 27: round.deck should return a deck object, so we can do something like: 

`expect(round.deck).to eq(deck)` (where `deck` is the deck object you defined on line 24)

spec/round_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+  it "can take turns" do
+    card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    card_2 = Card.new("The Viking spacecraft sent back to Earth photographs and reports about the surface of which planet?", "Mars", :STEM)
+    card_3 = Card.new("Describe in words the exact direction that is 697.5° clockwise from due north?", "North north west", :STEM)
+
+    deck = Deck.new([card_1, card_2, card_3])
+    round = Round.new(deck)
+    new_turn = round.take_turn("Juneau")
+
+    expect(round.current_card).to eq(card_1)
```
megstang, line 39: after we take a turn, we'll want the current card to be the next card in the deck, so we'll want to make sure this is card_2, not card_1.

spec/turn_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  it 'has a card' do
+    card = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    turn = Turn.new("Juneau", card)
+
+    expect(turn.card.question).to eq("What is the capital of Alaska?")
+    expect(turn.card.answer).to eq("Juneau")
+    expect(turn.card.category).to eq(:Geography)
+  end
```
megstang, line 28: for this spec, we mainly only need to test that turn.card returns the card object you initialized the object with. We don't need to check the attributes on that card object because we've already tested that in the Card spec
.

megstang (commented):
@kangkna I left a few comments for you to look over. Please let me know if you have any questions on the feedback I gave. I'll follow up by the end of the day tomorrow (via slack) with your scores.

PR #453, opened 2022-07-10
Carter Ball

cballrun:


lib/turn.rb
```diff
@@ -0,0 +1,26 @@
...
+  def initialize(guess, card)
+    @guess = guess
+    @card = card
+  end
+
+  def correct?
+    if @guess == card.answer
+      then true
+      else false
+    end
```
epintozzi, line 13: I'm not taking points off for this, but a more conventional way to write this if state would be like this:
```ruby
    if @guess == card.answer
      true
    else
      false
    end
```

lib/turn.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def feedback
+    if @guess == card.answer
+      then
```
epintozzi, line 18: We don't need to use `then` here. This should still work fine without it!

spec/turn_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+describe 'matches the turn guess to the card answer' do
+  it 'Turn #correct?' do
+      card = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+      turn = Turn.new("Juneau", card)
+
+      expect(turn.correct?).to eq(true)
+    end
+  end
```
epintozzi, line 37: It's a good idea to test all possible outcomes so I'd also add a test like this:
```ruby
    it 'Turn #correct? when false' do
      card = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
      turn = Turn.new("Sitka", card)

      expect(turn.correct?).to eq(false)
    end
```

spec/turn_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  describe 'gives the right feedback for a correct answer' do
+    it 'Turn #feedback' do
+        card = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+        turn = Turn.new("Juneau", card)
+
+        expect(turn.feedback).to eq('Correct!')
+      end
+    end
```
epintozzi, line 46: Another another here to test feedback when incorrect:
```ruby
      it 'Turn #feedback when incorrect' do
        card = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
        turn = Turn.new("Sitka", card)

        expect(turn.feedback).to eq('Incorrect.')
      end
```

spec/deck_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  describe 'has 3 cards' do
+    it 'Cards #count' do
+      card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+      card_2 = Card.new("The Viking spacecraft sent back to Earth photographs and reports about the surface of which planet?", "Mars", :STEM)
+      card_3 = Card.new("Describe in words the exact direction that is 697.5° clockwise from due north?", "North north west", :STEM)
+      cards = [card_1, card_2, card_3]
+      deck = Deck.new(cards)
+
+      expect(cards.count).to eq(3)
```
epintozzi, line 24: Since you wrote your own `count` method for `Deck` I would also add this here:
`expect(deck.count).to eq(3)`

This way the `count` method that you created gets tested too.

spec/deck_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    describe 'can return cards by category' do
+      it 'Cards #cards_in_category' do
+        card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+        card_2 = Card.new("The Viking spacecraft sent back to Earth photographs and reports about the surface of which planet?", "Mars", :STEM)
+        card_3 = Card.new("Describe in words the exact direction that is 697.5° clockwise from due north?", "North north west", :STEM)
+        cards = [card_1, card_2, card_3]
+        deck = Deck.new(cards)
+    
+        expect(deck.cards_in_category(:Geography).count).to eq(1)
+        expect(deck.cards_in_category(:STEM)).to eq([card_2, card_3])
```
epintozzi, line 36: It's best to be as explicit and consistent as possible, so I'd also add a test to verify that `:Geography` returns `card_1` and also that `:STEM` returns 2 cards.
```ruby
        expect(deck.cards_in_category(:Geography).count).to eq(1)
        expect(deck.cards_in_category(:Geography)).to eq([card_1])
        expect(deck.cards_in_category(:STEM).count).to eq(2)
        expect(deck.cards_in_category(:STEM)).to eq([card_2, card_3])
```

lib/deck.rb
```diff
@@ -0,0 +1,21 @@
...
+    end
+
+    def cards_in_category(category)
+      sorted_by_category = []
+      cards.select do |card|
+        if card.category == category
+          sorted_by_category << card
+        end
+    end
+  end
```
epintozzi, line 19: `select` will automatically return for us an array of all the objects that meet our condition (in this case `card.category == category`) so we don't need to define that part on our own. You could instead write this method as:
```ruby
    cards.select do |card|
      card.category == category
    end
  end
```

lib/round.rb
```diff
@@ -0,0 +1,59 @@
...
+        @correct_responses.length.to_f / @turns.length.to_f * 100
+    end
+
+    def percent_correct_by_category(category)
+        @correct_responses.select do |correct_response|
+            correct_response.card.category == category
+        end.length.to_f / @correct_responses.length.to_f * 100
+        
+
+
```
epintozzi, line 52: Remove extra lines

lib/round.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+   
+
```
epintozzi, line 55: Remove extra lines

lib/round.rb
```diff
@@ -0,0 +1,59 @@
...
+end
+
+
```
epintozzi, line 58: Remove extra lines

lib/round.rb
```diff
@@ -0,0 +1,59 @@
+class Round
```
epintozzi, line 1: The indentation in this class is a little wonky. Ruby convention is that each step should be indented two spaces. Most of these look to be 4 spaces.

lib/round.rb
```diff
@@ -0,0 +1,59 @@
...
+        attr_reader :deck, :turns, :number_correct
```
epintozzi, line 2: Because you have `number_correct` defined as a method on line 22 you do not need an `attr_reader` here

lib/round.rb
```diff
@@ -0,0 +1,59 @@
...
+        turn
+     end
+
+    def number_correct
+        turns.select do |turn|
+            if turn.correct? 
+                @correct_responses << turn
+            end
+        end
+        @correct_responses.length
```
epintozzi, line 28: `select` will return a collection of objects that match our condition so you do not need to manually put `turn` into `@correct_respones`. You are kind of combining two different ways of solving this problem. I would choose one or the other. My preference would be the first option.

Option 1:
```ruby
  def number_correct
    @correct_responses = turns.select do |turn|
      turn.correct?
    end
    @correct_responses.length
  end
```

Option 2
```ruby
  def number_correct
    turns.each do |turn|
      if turn.correct?
        @correct_responses << turn
      end
    end
    @correct_responses.length
  end
```

lib/round.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    def number_correct_by_category(category)
+        @correct_responses.select do |correct_response|
+            correct_response.card.category == category
+        end.length
```
epintozzi, line 34: Great use of `select` here!

lib/round.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    def percent_correct
+        turns.select do |turn|
+            if turn.correct? 
+                @correct_responses << turn
+            end
+        end
```
epintozzi, line 42: This code is identical to your `number_correct` method so I would just call that method here directly in place of lines 38-42

PR #454, opened 2022-07-10
Sandra Griffin 

SandyyMarie:


flashcard_runner.rb
```diff
@@ -0,0 +1,30 @@
...
+
+def start
+  filename = "./lib/cards.txt"
+  cards = CardGenerator.new(filename).cards
+  deck = Deck.new(cards)
+  round = Round.new(deck)
+
+  puts "Welcome! You're playing with #{deck.count} cards"
+  i = 1
+  deck.cards.each do |card|
```
epintozzi, line 15: Ruby arrays have a built in `each_with_index` method you might find handy here 😉

flashcard_runner.rb
```diff
@@ -0,0 +1,30 @@
...
+    given_answer = gets.chomp
+    round.take_turn(given_answer)
+    puts round.turns.last.feedback
+    i += 1
+  end
+
+  puts "****** Game over! ******"
+  puts "You had #{round.number_correct} correct guesses out of #{deck.count} for a total score of #{round.percent_correct}%."
+  puts "STEM - #{round.percent_correct_by_category(:STEM)}%"
+  puts "Geography - #{round.percent_correct_by_category(:TuringStaff)}%"
```
epintozzi, line 27: It would've been preferable to create a method that returns a list of all categories so that you don't need to hardcode this. It makes it more dynamic in case there are different categories in a deck.

lib/turn.rb
```diff
@@ -0,0 +1,23 @@
...
+              :card, 
+              :validity
+
+  def initialize(guess, current_card)
+    @guess = guess
+    @card = current_card
+  end
+
+  def correct?
+    @validity = @card.answer.include?(@guess)
```
epintozzi, line 13: I would consider updating this to `downcase` the `answer` and the `@guess` to try and allow a little more flexibility for the user.

spec/turn_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    expect(@turn.card).to be_instance_of(Card)
+  end
+
+  it "returns an answer" do
+    expect(@turn.guess).to eq("Juneau")
+  end
+
+  it "returns if correct or not" do
+    @turn.card
+    @turn.guess
```
epintozzi, line 24: These aren't really doing anything. You can take them out and the test will still work!

spec/turn_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    expect(@turn.correct?).to eq(true)
+  end
+
+  it "returns a string response of correct/incorrect" do
+    expect(@turn.guess).to eq("Juneau")
+    expect(@turn.correct?).to eq(true)
+    expect(@turn.feedback).to eq("Correct!")
+  end
+
+  it "will handle incorrect guesses" do
```
epintozzi, line 34: Great work considering both correct and incorrect guesses in your tests!

spec/deck_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  it "can return the count of cards in deck" do
+    expect(@deck.count).to eq(3)
+  end
+
+  it "can return the cards of a given deck category" do
+    expect(@deck.cards_in_category(:STEM).count).to eq(2)
+    expect(@deck.cards_in_category(:Geography).count).to eq(1)
+    expect(@deck.cards_in_category("Pop Culture")).to eq([])
```
epintozzi, line 24: I would also add the following to explicitly ensure the correct cards are returned.
```ruby
    expect(@deck.cards_in_category(:STEM)).to eq([@card_2, @card_3])
    expect(@deck.cards_in_category(:Geography)).to eq([@card_1])
```

lib/deck.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def cards_in_category(category)
+    return_array = []
+    @cards.each do |card|
+      if card.category == category
+        return_array << card
+      end
+    end
+    return_array
+  end
```
epintozzi, line 21: You can also use a `select` or a `find_all` here:
```ruby
    @cards.select do |card|
      card.category == category
    end
```

lib/round.rb
```diff
@@ -0,0 +1,69 @@
+class Round
+  attr_reader :deck, 
+              :turns, 
+              :current_turn, 
+              :number_correct, 
+              :correct_qs, 
+              :curr_card
```
epintozzi, line 7: You don't need this `attr_reader` because you never access `@curr_card` outside of this class so you can call `@curr_card` directly.

lib/round.rb
```diff
@@ -0,0 +1,69 @@
...
+    @correct_qs = []
+  end
+
+  def current_card
+    @curr_card = @deck.cards[@current_turn]
+  end
+
+  def take_turn(guess_string)
+   if @current_turn > @deck.count
+      return  "you have reached the limit of questions"
```
epintozzi, line 23: I'm not sure I see where this ever gets used. The counter in your runner would prevent this from ever happening, I think. You can probably get rid of this `if` statement and just start this method with `@new_turn = Turn.new....`

lib/round.rb
```diff
@@ -0,0 +1,69 @@
...
+
+ def number_correct_by_category(category_choice)
+   cat_correct = 0
+   @correct_qs.each do |card|
+     if card.category == category_choice
+       cat_correct += 1
+     end
+   end
+   cat_correct
+ end
```
epintozzi, line 49: We can use the built in `count` enumerator here instead of having to set up our own counter with `each`. This could be re-written as:
```ruby
 def number_correct_by_category(category_choice)
   @correct_qs.count do |card|
     card.category == category_choice
   end
 end
```

lib/round.rb
```diff
@@ -0,0 +1,69 @@
...
+   (@number_correct.to_f / @deck.count.to_f * 100).to_i
+ end
+
+ def percent_correct_by_category(category_choice)
+   num_cat = 0
+   @deck.cards.each do |card|
+     if card.category == category_choice
+       num_cat += 1
+     end
+   end
```
epintozzi, line 61: You already have a method in `Deck` for returning `cards_in_category` so I would use that here and then just `count`. Something like:
`num_cat = @deck.cards_in_category(category_choice).count`

lib/card_generator.rb
```diff
@@ -0,0 +1,22 @@
+# require './card'
+
+class CardGenerator
+  attr_reader :card_group
```
epintozzi, line 4: We don't need this `attr_reader` since `card_group` is not called outside of this class.

lib/card_generator.rb
```diff
@@ -0,0 +1,22 @@
...
+   @card_group = [] 
+   if File.exist?(filename) 
+     File.open(filename).each do |line|
+       answer_array = line.split(",")
+       new_card = Card.new(answer_array[0],answer_array[1],answer_array[2].strip)
+       @card_group << new_card
+     end
+   else
+    p "File not read"
+   end
```
epintozzi, line 16: I like that you've considered this edge case of what happens if a user provides a bad file. Nice job!

flashcard_runner.rb
SandyyMarie: Ahhh I love this! never fails to amaze me how many enumerable are available in ruby

PR #455, opened 2022-07-10
Ashley Turner

ashuhleyt:
Hi! Last week was so hard for me. I spoke with Megan a bit on Friday, and took her advice to spend Saturday rewatching the lessons again to better help the content stick with me. I did what I was able to do Sunday on the project, and, it is starting to make a lot more sense to me now. And after finally getting a mentor and being able to meet with him Saturday morning, I am starting to feel a lot better about my capabilities, and where I am at mentally. I felt like I was solely getting overwhelmed with the intense amount of new content we were getting, so rewatching them with a break from more content was helpful. I also ran into an issue trying to commit my work. I do have a card_spec, I promise, but was unable to get it to be pushed up to github, even after consulting with a few cohort mates. I appreciate the patience you guys have shown, and am proud of myself for even being able to complete the work I did here. Thanks.

### DIFF
--- a/lib/card.rb
+++ b/lib/card.rb
@@ -1,0 +1,10 @@
+class Card
+  attr_reader :question, :answer, :category
+
+
+   def initialize(question, answer, category)
+    @question = question
+    @answer = answer
+    @category = category
+  end
+end
--- a/lib/deck.rb
+++ b/lib/deck.rb
@@ -1,0 +1,21 @@
+require './lib/deck'
+require './lib/card'
+require 'pry'
+
+class Deck
+  attr_reader :cards
+  def initialize(cards)
+    @cards = cards
+  end
+
+  def count
+    @cards.count
+  end
+
+  def cards_in_category(category)
+    @cards.find_all do |card|
+      card.category == category
+    end
+  end
+
+end
--- a/lib/turn.rb
+++ b/lib/turn.rb
@@ -1,0 +1,25 @@
+require './lib/turn'
+require './lib/card'
+require './lib/deck'
+require './lib/round'
+
+class Turn
+    attr_reader :cards, :turns
+
+  def initialize(cards, turns)
+    @cards = card
+    @turns = turn
+  end
+
+  def guess
+    @turns.guess #need this to return guess
+  end
+
+  def number_correct?
+    @cards == correct
+  end
+
+  def feedback
+    @cards.none? #unsure of what method to call here
+  end
+end
--- a/spec/turn_spec.rb
+++ b/spec/turn_spec.rb
@@ -1,0 +1,21 @@
+require './lib/turn'
+require './lib/card'
+require './lib/deck'
+require './lib/round'
+
+RSpec.describe Turn do
+
+  card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+  card_2 = Card.new("The Viking spacecraft sent back to Earth photographs and reports about the surface of which planet?", "Mars", :STEM)
+  card_3 = Card.new("Describe in words the exact direction that is 697.5° clockwise from due north?", "North north west", :STEM)
+#this section is wrong. I believe I need to use each here, but, really unsure of how to use it.
+end
+
+it 'exists' do
+  expect(turn).to be_instance_of(Turn)
+end
+
+it 'has a guess' do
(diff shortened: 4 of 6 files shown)


lib/card.rb
```diff
@@ -0,0 +1,10 @@
+class Card
+  attr_reader :question, :answer, :category
+
+
+   def initialize(question, answer, category)
+    @question = question
+    @answer = answer
+    @category = category
+  end
+end
```
megstang, line 10: This looks great! For syntax sake, just do one blank line between the attr_readers and the initialize method.

lib/deck.rb
```diff
@@ -0,0 +1,21 @@
+require './lib/deck'
+require './lib/card'
+require 'pry'
```
megstang, line 3: pry is a debugging tool. So, for that reason, I'd just require it in your tests. It doesn't hurt to have it here.. but you don't need it.

lib/deck.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def count
+    @cards.count
+  end
+
+  def cards_in_category(category)
+    @cards.find_all do |card|
+      card.category == category
+    end
```
megstang, line 18: Nice use of `find_all` here!

lib/turn.rb
```diff
@@ -0,0 +1,25 @@
+require './lib/turn'
+require './lib/card'
+require './lib/deck'
+require './lib/round'
+
+class Turn
+    attr_reader :cards, :turns
```
megstang, line 7: indent this back a bit, so it's in line with where `def` is on your method declarations below.

lib/turn.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def initialize(cards, turns)
+    @cards = card
+    @turns = turn
+  end
```
megstang, line 12: Go ahead back to the interaction pattern and take a look at the Turn Class interaction pattern. It shows that when a Turn object gets initialized, it takes in two arguments (which you have), but, it should be taking in a guess and the card that's being guessed on.

megstang: You can initialize this class with the card and the guess, and then create attr_readers to be able to read the card and guess that get passed in.

lib/turn.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def guess
+    @turns.guess #need this to return guess
+  end
```
megstang, line 16: No need for this method. Based on my feedback above, in order to have a method `guess` that will return the guess that the Turn object was initialized with, you'll need to create an attr_Reader to read that value.

lib/turn.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def number_correct?
+    @cards == correct
+  end
```
megstang, line 20: this method is for a different class (Round)

lib/turn.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def feedback
+    @cards.none? #unsure of what method to call here
+  end
```
megstang, line 24: In this method, we want to check if the guess is equal to the cards answer. If it is, then this method should return appropriate feedback like "Correct!" and "Wrong!"

spec/turn_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+end
+
+it 'exists' do
+  expect(turn).to be_instance_of(Turn)
+end
+
+it 'has a guess' do
+  expect(turn.guess).to eq("Juneau") #I know that I need to use method chaining here, but, uncertain
+  #if this is correct
+end
```
megstang, line 21: Think back, or reference, your Deck spec. In each of those `it` blocks, you created card objects to test out. Similarly here, if you need objects to help test certain methods, create those in the it blocks. At the moment, you have a describe block with three cards, and then those it blocks exist outside of the describe block. In order for your it blocks to know the objects that you're trying to test, we need to move them into the it blocks.

megstang (commented):
@ashuhleyt Please let me know if you have any questions on the comments I put on your code. I'm glad you were able to take the weekend to practice the concepts taught in the week 1 lessons, and that you're feeling more confident after doing so. I'll follow up in slack with scores by the end of the day tomorrow.

PR #457, opened 2022-07-10
Noah van Ekdom Flash Cards

noahvanekdom:


lib/deck.rb
```diff
@@ -0,0 +1,22 @@
+class Deck
+    attr_reader :cards
+
+  def initialize(cards)
+    @cards = cards
+  end
+
+  def count
+    cards.count
```
megstang, line 9: Instead of calling `cards.count`, call to the instance variable `@cards.count`. When you call to `cards` it is actually calling to the attr_Reader on line 2, which then returns the instance variable `@cards`. To optimize your code, cut out the middle man and call directly to `@cards`

lib/deck.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def cards_in_category(category)
+    cards.find_all do |card|
+        card.category == category
+        end
+    end
```
megstang, line 16: pull these two ends in one more tab.

megstang, line 16: Nice use of `find_all`. Once again, reference instance variables when you can instead of their methods that got generated from the attr_readers.

lib/round.rb
```diff
@@ -0,0 +1,47 @@
...
+        @turns << new_turn
+        if new_turn.correct?
+            @number_correct += 1
+        end
+        new_turn
+    end
+    
+    def number_correct
+        @number_correct
+    end
```
megstang, line 27: You could make an attr_reader for this.

lib/round.rb
```diff
@@ -0,0 +1,47 @@
+class Round
+    attr_reader :deck
+    attr_reader :turns
```
megstang, line 3: Instead of having these on two lines, you can put them on the same line by separating the symbols by columns. `attr_reader :deck, :turns`

lib/round.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+    def take_turn(guess)
+        new_turn = Turn.new(guess, current_card)
+        @turns << new_turn
+        if new_turn.correct?
+            @number_correct += 1
+        end
+        new_turn
+    end
```
megstang, line 23: 👍 Nice work here

lib/round.rb
```diff
@@ -0,0 +1,47 @@
...
+    
+    def number_correct
+        @number_correct
+    end
+    
+    def number_correct_by_category(category)
+        @turns.count do |turn|
+            turn.card.answer == turn.guess && turn.card.category == category
```
megstang, line 31: You already have a method in the `Turn` class called `correct?` that you can lean on for the first half of this condition. `turn.correct? && turn.card.category == category`

lib/round.rb
```diff
@@ -0,0 +1,47 @@
...
+    def percent_correct
+        @number_correct.to_i / @turns.count.to_i * 100.0
+    end
+
+    def percent_correct_by_category(category)
+        total_turns_by_category = @turns.count do |turn|
+            turn.card.category == category 
+        end
+        100*number_correct_by_category(category).to_i / total_turns_by_category.to_i
+    end
```
megstang, line 45: If we were to assume that this method gets called at the end of the game, when all cards in the deck have been guessed on, you could use the `cards_in_category` method in your deck class to get the number of cards in that category within the deck. I do like that the way you have your method written does make this calculation dynamic for however many cards have been guessed upon at any given moment. So, truly no preference on my end, just wanted to point it out in case you didn't think of that ;)

lib/turn.rb
```diff
@@ -0,0 +1,24 @@
...
+        @guess = guess
+        @card = card
+    end
+
+    def correct?
+        @guess == card.answer
+    end
+
+    def response
+        if correct? == true
```
megstang, line 18: nice job using the method above so that you can keep your code DRY (Don't Repeat Yourself)

runner.rb
```diff
@@ -0,0 +1,39 @@
...
+  puts ("----------")
+
+  card_number = 0
+  deck.cards.each do |card|
+    card_number +=1
+    puts card.question
+    guess = gets.chomp.to_s
+    round.take_turn(guess)
+    round.turns.last.correct?
+    puts round.turns.last.feedback
```
megstang, line 25: I think this line is the reason you were getting that error when trying to run the game. The method that returns feedback in your turn class is actually named `response`. So this line should read `puts round.turns.last.response`

runner.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    if card_number == deck.count
+      puts "Game Over!"
+      puts "You had #{round.number_correct} out of #{deck.count} for a total score of #{round.percent_correct}%."
+      puts "STEM - #{round.percent_correct_by_category(:STEM)}% correct"
+      puts "Geography - #{round.percent_correct_by_category(:Geography)}% correct"
+      puts "Animals - #{round.percent_correct_by_category(:Animals)}%"
```
megstang, line 32: Your card_4's category is :Popculture, so this line of code will error our because in your percent_correct_by_category, you'll be dividing by 0 which is a no no in math

megstang (commented):
@noahvanekdom Please look through these comments and let me know if you have any questions. I'll follow up in slack by the end of the day tomorrow with your scores.

PR #458, opened 2022-07-10
Taryn Orlemann

torlemann:


lib/turn.rb
```diff
@@ -0,0 +1,21 @@
...
+    @guess = guess
+    @card = card
+  end
+
+  def correct?
+    @guess == @card.answer
+  end
+
+  def feedback
+    if correct?
```
abdulredd, line 15: Great job using the previously defined `correct?` method to your advantage and not re-writing the same line over again.

lib/deck.rb
```diff
@@ -0,0 +1,25 @@
...
+  # def cards_in_category(category)
+  #   card_category = []
+  #   @cards.each do |card|
+  #     card_category << card if card.category == category
+  #   end
+  #   card_category
+  # end
+
+  def cards_in_category(category)
+    @cards.select do |card|
```
abdulredd, line 21: I like that you chose the `select` method here instead of relying on `each`.

lib/round.rb
```diff
@@ -0,0 +1,41 @@
...
+  def take_turn(guess)
+    @turn = Turn.new(guess, current_card)
+    @turns << @turn
+    @deck.cards.rotate!
+    @turn
+    #change the current card to the next card in the deck
+  end
+
+  def number_correct
+    @turns.count do |turn|
```
abdulredd, line 23: I love this, most people don't realize you can pass a block to the `count` method. This keeps your code nice and clean.

PR #459, opened 2022-07-11
Kristen Nestler

knestler:


lib/turn.rb
```diff
@@ -0,0 +1,29 @@
...
+    def correct? 
+        if @guess == @card.answer
+          true
+        else
+            false
+        end
+    end
+
+    def feedback
+        if @guess == @card.answer
```
abdulredd, line 23: This line here `@guess == @card.answer` looks familiar line 15, so how can we use that method to help us not have to duplicate this line. Let's think about the feedback method in simple english terms, we would say ... "If the answer is correct, then return "Correct", if not return "Incorrect"". And aren't we lucky that we already have a method that checks if the answer is correct or not, so we can re-write this feedback method like so:

```ruby
def feedback
  if correct? # <=== Notice how we use the already defined method 'correct?' to our advantage
    "Correct!"
  else
    "Incorrect."
  end
end
```

lib/round.rb
```diff
@@ -0,0 +1,61 @@
+require './lib/deck'
```
abdulredd, line 1: You don't need this `require './lib/deck'` because you are not instantiating a new `Deck` object in the `Round` class, the same thing goes for `require './lib/card'`, the only object that you are instantiating in the class a `Turn` object on line 22, so the `require './lib/turn'` is good.

lib/round.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    attr_reader :deck, :turns
+
+    def initialize(deck)
+        @deck = deck
+        @turns = []
+    end
+
+    def current_card
+        @deck.cards.first
```
abdulredd, line 16: While this piece of code would work correctly for the first turn, because of how it is written (always returning the first card in the deck), every subsequent turn will always be the first card in the deck so all other answers would end up being incorrect no matter what.

The solution to this, if you wanted to keep `@deck.cards.first` would be to rotate the array of cards so that the first card goes to the back of the deck and all other cards move up by one. This was the first card will always be the next card in the deck after the rotation.

lib/round.rb
```diff
@@ -0,0 +1,61 @@
...
+        correct = 0
+        @turns.each do |turn_new|
+            if turn_new.correct?
+                correct += 1
+            end
+        end
+        return correct
+    end
+
+    def number_correct_by_category(category)
```
abdulredd, line 36: This method has the right idea, but wouldn't currently work as written. You could definitely use the `@turns` array to determine the number of correct cards by category, but you would first need to determine if the turn is correct and the turn card is of the right category (which is what you are currently doing on line 38). After which you should be able to get the count of cards from the `find_all` call, as `find_all` doesn't return a number, but the actual array of elements that pass the block condition.

spec/turn_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+RSpec.describe Turn do
+
+  before do
+    @card = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+    @turn = Turn.new("Juneau", @card)
+  end 
+  
+  it 'card exists' do
+    expect(@card).to be_instance_of(Card)
```
abdulredd, line 13: Didn't we already run this test in the `card_spec.rb` file? Maybe instead of creating an expectation for `@card` we could make one for `@turn.card` to make sure that it is an instance of `Card`?

spec/turn_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  it 'turn exists' do
+    expect(@turn).to be_instance_of(Turn)
+  end
+
+  it 'is guess' do
+    expect(@turn.guess).to eq("Juneau")
+  end
+
+  it 'evaluated if correct' do
+    expect(@turn.correct?).to eq(true)
```
abdulredd, line 25: When you are testing a boolean method like `correct?` you should not only test if it returns `true` but also if you can also return `false` so that was you can make sure that it works under both conditions.

The same thing for the `feedback` method below.

spec/deck_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    before do 
+        @card_1 = Card.new("What is the capital of Alaska?", "Juneau", :Geography)
+        @card_2 = Card.new("The Viking spacecraft sent back to Earth photographs and reports about the surface of which planet?", "Mars", :STEM)
+        @card_3 = Card.new("Describe in words the exact direction that is 697.5° clockwise from due north?", "North north west", :STEM)
+        @cards = [@card_1, @card_2, @card_3]
+    end
+
+
+    it 'deck exists' do
+        deck = Deck.new(@cards)
```
abdulredd, line 18: It looks like every test in this spec requires `deck = Deck.new(@cards)`... would it have been better to just place this line in the `before` block, thus not having to repeat it for every test?

spec/round_spec.rb
```diff
@@ -0,0 +1,101 @@
+require './lib/round'
+require './lib/deck'
+require './lib/turn'
+require './lib/card'
+require 'pry'; #binding.pry
+
+RSpec.describe Round do
```
abdulredd, line 7: I think you are missing the check for the `initialize` method and to make sure that you are actually successfully making an instance of `Round`.

PR #642, opened 2025-02-11
Stephanie Guerrero 

Sguerrero0991:


Sguerrero0991:
I did not complete the project but I did as much as I could using the resources that I had. I needed help for most of the project and I felt like I didn't really have the knowledge needed to figure this out on my own. I don't think I could have gotten through most of this without the help of my instructors, my mentors, and my rocks, chat GBT and googling. I went over our lessons various times and still felt pretty lost through most of this.

